PR #4, opened 2020-12-02
Wr/data bulk commands @W-8449681@

WillieRuemmele:
### What does this PR do?
adds data:bulk:* commands
@W-8449681@

* fix https://github.com/forcedotcom/cli/issues/423 / W-7837787
* fix https://github.com/forcedotcom/cli/issues/635


amphro (changes requested):
I'll look at this more in-depth when some of the high-level comments are addressed. I was talking too @shetzel about this. If we are publishing a library I want us to make sure we are happy with the API (classes and method signatures) and they all make sense. Otherwise, if we don't want to refactor too much and more stuff over directly, I would rather this code go in the plugin and we refactor to the library later. OR we just have to make sure we publish a beta (0.X.X) release and rrefactorr everything before we make it a major, but that feels "harder".

packages/plugin-data/src/commands/force/data/bulk/delete.ts
```diff
@@ -0,0 +1,69 @@
...
+      description: messages.getMessage('flags.sobjecttype'),
+      required: true,
+    }),
+    wait: flags.number({
+      char: 'w',
+      description: messages.getMessage('flags.wait'),
+      min: 0,
+    }),
+  };
+  public org!: Org;
```
shetzel, line 38: This shouldn't be necessary since it's on SfdxCommand.

packages/data/src/batcher.ts
```diff
@@ -0,0 +1,319 @@
...
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { ReadStream } from 'fs';
+import { Connection, Logger, Messages, SfdxError } from '@salesforce/core';
+import parse = require('csv-parse');
+import { Batch, BatchInfo, BatchResultInfo, JobInfo } from 'jsforce';
+import { UX } from '@salesforce/command';
```
shetzel, line 11: I don't think this library should use anything from `@salesforce/command`.  In fact, if it's a dependency listed in package.json we should remove it.

Also, I think we should always use `fs` from `@salesforce/core` since it protects us against some problems we've run into with the native fs lib.

shetzel (changes requested):
I think I have the same concerns as Thomas particularly about the library API.  There is a lot there that doesn't look like a good direction to take.  Maybe we should move it all to the plugin and think about the public API as a dedicated work item.  I'm leaning towards doing the same thing with the data:tree code.

packages/data/src/batcher.ts
```diff
@@ -0,0 +1,319 @@
...
+  createdDate: string;
+  systemModStamp: string;
+  numberRecordsProcessed: string;
+  numberRecordsFailed: string;
+  totalProcessingTime: string;
+  apiActiveProcessingTime: string;
+  apexProcessingTime: string;
+};
+
+export class Batcher {
```
mdonnalley, line 39: As we put stuff into this data library we need to be thinking about how people can use it for their own purposes. As it is now, I think it's a little unclear how people would use this or why they would use it.

So I'm wondering if there's a way to make this a little more user friendly. My first inclination is to rename it to `Job` and have it include all the logic for creating the job that's currently in the plugin. That way the commands merely wrap around this class.

Let me know if that doesn't make sense

packages/data/test/.eslintrc.js
```diff
@@ -12,6 +12,8 @@ module.exports = {
   rules: {
     // Allow assert style expressions. i.e. expect(true).to.be.true
     'no-unused-expressions': 'off',
+    '@typescript-eslint/ban-ts-ignore': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
```
mdonnalley, line 16: revert this

packages/data/src/batcher.ts
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  public static bulkBatchStatus(
+    summary: JobInfo | BatchInfo,
+    ux: UX,
+    results?: BatchResultInfo[],
+    batchNum?: number,
+    isJob?: boolean
+  ): void {
+    ux.log('');
```
mdonnalley, line 73: Is there any way to avoid passing in `ux`? Again, thinking about this from a usability perspective, people may not be using this library in the context of a plugin so they wouldn't have and wouldn't need the `ux` class

mdonnalley (changes requested):

packages/plugin-data/src/batcher.ts
```diff
@@ -0,0 +1,303 @@
...
+   */
+  private async waitForCompletion(
+    newBatch: Batch,
+    batchesCompleted: number,
+    overallInfo: boolean,
+    batchNum: number,
+    totalNumBatches: number,
+    waitMins: number
+  ): Promise<void> {
+    const logger: Logger = await Logger.child(this.constructor.name);
```
shetzel, line 224: I think we should create the logger synchronously in the constructor, or a new one will be created every time the function is called.

shetzel (approved):

packages/data/test/.eslintrc.js
mdonnalley: This still needs to be reverted

packages/plugin-data/src/batcher.ts
mdonnalley: Agree with Steve

mdonnalley (approved):

packages/plugin-data/src/batcher.ts
```diff
@@ -55,10 +55,10 @@ export class Batcher {
     jobId: string,
     doneCallback?: (...args: [{ job: JobInfo }]) => void
   ): Promise<JobInfo> {
-    const job: jsforceJob = this.conn.bulk.job(jobId);
+    const job: Job = this.conn.bulk.job(jobId);
     const jobInfo: JobInfo = await job.check();
 
-    this.bulkBatchStatus(jobInfo);
+    this.bulkStatus(jobInfo, undefined, undefined, true);
```
shetzel, line 61: JSYK: A way to avoid passing undefined is to change the method signature to take a config object (best way) or possibly to use default values.

shetzel (approved):

mshanemc:
qa notes:
I ran a batch, then changed the job Id by one letter.  The commandName, name, and stack trace on the error are different.  (bin/run is the new plugin, sfdx is the old toolbelt)

Screen Shot 2021-01-22 at 6 31 34 PM

mshanemc:
qa notes:
I ran a batch:upsert with this command, expecting it to wait 1 min.  `../bin/run force:data:bulk:upsert -f data/Reporting_2__c.csv -s Reporting_2__c -i ID --json -w 1`

It never returned at all...seemed stuck.  I eventually had to ctrl-c it

mshanemc:
qa notes:
I re-ran that command with the -w 1 again.  This time it did return, but returns no json (known bug in toolbelt, too).  Would prefer to return error code 69 for timeouts and return the jobid so user can get past it.

PR #14, opened 2020-12-11
Moves the data tree commands, messages, APIs, and tests from salesfor…

shetzel:
…ce-alm to @salesforce/plugin-data

### What does this PR do?
Adds `force:data:tree:export` and `force:data:tree:import` commands
Adds messages
Adds API classes that are not ready to be moved to `@salesforce/data`
Adds unit tests

@W-8449687@

### DIFF
--- a/packages/plugin-data/package.json
+++ b/packages/plugin-data/package.json
@@ -28,15 +28,7 @@
                 "description": "fetch records using SOQL"
               },
               "tree": {
-                "description": "import/export records using the tree API",
-                "subtopics": {
-                  "import": {
-                    "description": "import records using the tree API"
-                  },
-                  "export": {
-                    "description": "export records using the tree API"
-                  }
-                }
+                "description": "import/export records using the tree API"
               }
             }
           }
--- a/packages/plugin-data/package.json
+++ b/packages/plugin-data/package.json
@@ -28,7 +28,15 @@
                 "description": "fetch records using SOQL"
               },
               "tree": {
-                "description": "manipulate records using the tree API"
+                "description": "import/export records using the tree API",
+                "subtopics": {
+                  "import": {
+                    "description": "import records using the tree API"
+                  },
+                  "export": {
+                    "description": "export records using the tree API"
+                  }
+                }
               }
             }
           }
--- a/packages/plugin-data/src/api/data/tree/exportApi.ts
+++ b/packages/plugin-data/src/api/data/tree/exportApi.ts
@@ -368,8 +368,7 @@ export class ExportApi {
     });
 
     if (!result) {
-      const metadataName: string = metadata.name;
-      throw new SfdxError(`Unable to find relationship field name for ${metadataName}`);
+      throw new SfdxError(`Unable to find relationship field name for ${metadata.name as string}`);
     }
 
     return result;
@@ -411,8 +410,7 @@ export class ExportApi {
     });
 
     if (!result) {
-      const metadataName: string = metadata.name;
-      throw new SfdxError(`Unable to find relation for ${metadataName}`);
+      throw new SfdxError(`Unable to find relation for ${metadata.name as string}`);
     }
 
     return result;
--- a/packages/plugin-data/src/api/data/tree/exportApi.ts
+++ b/packages/plugin-data/src/api/data/tree/exportApi.ts
@@ -1,0 +1,607 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+
+import * as path from 'path';
+import { fs, Logger, Messages, Org, SfdxError } from '@salesforce/core';
+import { AnyArray, AnyJson, get, getNumber, getString, has, isArray } from '@salesforce/ts-types';
+import { UX } from '@salesforce/command';
+import { sequentialExecute } from './executors';
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/plugin-data', 'exportApi');
+
+const DATA_PLAN_FILENAME_PART = '-plan.json';
+
+const describe: any = {}; // holds metadata result for object type describe calls
+
+interface DataPlanPart {
+  sobject: string;
+  saveRefs: boolean;
+  resolveRefs: boolean;
+  files: string[];
+}
+
+export interface ExportConfig {
+  query: string;
+  outputDir?: string;
+  plan?: boolean;
+  prefix?: string;
+}
+
+/**
+ * Exports data from an org into sObject tree format.
+ */
+export class ExportApi {
+  private logger: Logger;
+
+  private objectTypeRegistry: any = {}; // registry for object type data plan descriptor
+  private referenceRegistry: any = {}; // registry of object type-specific id-to-ref mappings
+  private typeRefIndexes: any = {}; // registry for object type-specific ref counters
+
+  private config!: ExportConfig;
+
+  public constructor(private readonly org: Org, private readonly ux: UX) {
+    this.logger = Logger.childFromRoot(this.constructor.name);
+  }
+
+  /**
+   * Invokes the provided SOQL query against a target Org.  Results
+   * are converted into SObject Tree format.
+   *
+   * @param config
+   */
+  public async export(config: ExportConfig): Promise<AnyJson> {
+    this.config = this.validate(config);
+
+    const { outputDir, plan, query } = this.config;
+
+    if (outputDir) {
+      this.setupOutputDirectory(outputDir);
+    }
+
+    let queryResults: any;
+    try {
+      queryResults = await this.org.getConnection().query(query);
+    } catch (err) {
+      if (err.name === 'MALFORMED_QUERY') {
+        const errMsg = messages.getMessage('soqlMalformed');
+        const errMsgAction = messages.getMessage('soqlMalformedAction');
+        throw new SfdxError(errMsg, 'MalformedQuery', [errMsgAction]);
+      } else {
+        throw err;
+      }
+    }
+
+    const sobjectTree = await this.processQueryResults(queryResults);
+
+    if (!sobjectTree.records?.length) {
+      return sobjectTree;
+    }
+    if (plan) {
+      return this.generateDataPlan(sobjectTree);
+    } else {
+      const fileName = `${Object.keys(this.objectTypeRegistry).join('-')}.json`;
+      return this.writeDataFileSync(fileName, sobjectTree);
+    }
+  }
+
+  //   P R I V A T E   M E T H O D S
+
+  /**
+   * Ensures a valid query is defined in the export configuration,
+   * which can be either a soql query or a path to a file containing
+   * a soql query.
+   *
+   * @param config - The export configuration.
+   */
+  private validate(config: ExportConfig): ExportConfig {
+    if (!config.query) {
+      throw SfdxError.create('@salesforce/plugin-data', 'exportApi', 'queryNotProvided');
+    }
+
+    const filepath = path.resolve(process.cwd(), config.query);
+    if (fs.existsSync(filepath)) {
+      config.query = fs.readFileSync(filepath, 'utf8');
+
+      if (!config.query) {
+        throw SfdxError.create('@salesforce/plugin-data', 'exportApi', 'queryNotProvided');
+      }
+    }
+
+    config.query = config.query.toLowerCase().trim();
+    if (!config.query.startsWith('select')) {
+      throw SfdxError.create('@salesforce/plugin-data', 'exportApi', 'soqlInvalid', [config.query]);
+    }
+
+    return config;
+  }
+
+  private setupOutputDirectory(outputDir: string): void {
+    try {
+      fs.mkdirpSync(outputDir);
+    } catch (err) {
+      // It is ok if the directory already exist
+      if (err.name !== 'EEXIST') {
+        throw err;
+      }
+    }
+  }
+
+  // Process query results generating SObject Tree format
+  private async processQueryResults(recordList: any): Promise<any> {
+    await this.recordObjectTypes(recordList);
+
+    const { plan, query } = this.config;
+
+    const processedRecordList = await this.processRecordList(recordList);
+    // log record count; warn if > 200 and !options.plan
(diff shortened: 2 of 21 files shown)


packages/plugin-data/src/commands/force/data/tree/export.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Messages, Org } from '@salesforce/core';
+import { ExportApi } from '../../../../lib/data/tree/exportApi';
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/plugin-data', 'tree.export');
+
+export default class Export extends SfdxCommand {
+  public static readonly description = messages.getMessage('description');
+  public static readonly examples = messages.getMessage('examples').split(os.EOL);
+  public static readonly requiresProject = false;
```
WillieRuemmele, line 21: Is this defaulted to false? I thought most things were

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+export class ExportApi {
+  private logger: Logger;
+
+  private objectTypeRegistry: any = {}; // registry for object type data plan descriptor
+  private referenceRegistry: any = {}; // registry of object type-specific id-to-ref mappings
+  private typeRefIndexes: any = {}; // registry for object type-specific ref counters
+
+  private config!: ExportConfig;
+
+  public constructor(private readonly org: Org, private readonly ux: UX) {
```
WillieRuemmele, line 49: Why are we passing in org and ux if they're not used in the constructor?

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+  public async export(config: ExportConfig): Promise<AnyJson> {
+    this.config = this.validate(config);
+
+    const { outputDir, plan, query } = this.config;
+
+    if (outputDir) {
+      this.setupOutputDirectory(outputDir);
+    }
+
+    let queryResults: any;
```
WillieRuemmele, line 68: how did you get this `any` here? I think there's a `QueryResults` type you can import from jsforce or maybe from core

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+    }
+    return this.finalApplyRefs(processedRecordList.records);
+  }
+
+  // Register object types and type hierarchy for plan generation
+  private async recordObjectTypes(recordList: any): Promise<any> {
+    const records = recordList.records;
+
+    if (!records.length) {
+      // TODO: should be on the command
```
WillieRuemmele, line 157: why don't we do this now?

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+
+  private async processRecordList(recordList: any, parentRef?: any): Promise<any> {
+    // holds transformed sobject tree
+    const sobjectTree = { records: [] };
+
+    // visit each record in the list
+    const processRecordsFn = (record: any) => (): Promise<any> => this.processRecords(parentRef, record, sobjectTree);
+    const recordFns = recordList.records.map((record: any) => processRecordsFn(record));
+
+    // TODO: do we really need sequentialExecute?  Can't we just use Promise.all()?
```
WillieRuemmele, line 203: can't we? I feel like this is the time to fix all of these TODOs and make the plugin as pristine as possible

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+    return Promise.resolve().then(() => {
+      const field = record[key];
+
+      // skip attributes and id.  Data import does not accept records with IDs.
+      if (key === 'attributes' || key === 'Id') {
+        // If this is an attributes section then we need to add an object reference
+        this.saveRecordRef(record, objRefId);
+      } else {
+        return this.loadMetadata(record.attributes.type)
+          .then((metadata) => {
```
WillieRuemmele, line 264: Anyway we can refactor this to get rid of the promise chain?

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+            return true;
+          }
+        }
+      }
+
+      return false;
+    });
+
+    if (!result) {
+      const metadataName: string = metadata.name;
```
WillieRuemmele, line 371: absolute nit: but why have a variable for this? just print `${metadata.name}`

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+          return true;
+        }
+      }
+
+      return false;
+    });
+
+    if (!result) {
+      const metadataName: string = metadata.name;
+      throw new SfdxError(`Unable to find relation for ${metadataName}`);
```
WillieRuemmele, line 415: same NIT as above

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+
+    return dataPlanPart;
+  }
+
+  /**
+   * This method is used as a second pass to establish references that couldn't be determined
+   * in the initial pass done by processRecordList. It looks for relationship fields that
+   * contain an id.
+   */
+  private finalApplyRefs(sobjectTree: any): { records: any } {
```
WillieRuemmele, line 534: probably just me but I feel like this method could use a more descriptive name

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -0,0 +1,607 @@
...
+      fileName = path.join(outputDir, fileName);
+    }
+
+    if (has(jsonObject, 'records') && isArray(jsonObject.records)) {
+      recordCount = this.countRecords(jsonObject.records);
+    }
+
+    fs.writeFileSync(fileName, JSON.stringify(jsonObject, null, 4));
+
+    // TODO: move this to the command
```
WillieRuemmele, line 602: let's do it

packages/plugin-data/src/api/data/tree/importApi.ts
```diff
@@ -0,0 +1,496 @@
...
+  private sobjectUrlMap: Map<string, string>;
+  private schemaValidator: SchemaValidator;
+  private sobjectTypes: any = {}; // hash of sobject { referenceId: type }
+
+  private config!: ImportConfig;
+
+  // A JsonArray but we need to provide a better type to work with it properly.
+  private importPlanConfig: any;
+
+  public constructor(private readonly org: Org) {
```
WillieRuemmele, line 78: why are we passing in org if we're not using it?

packages/plugin-data/src/api/data/tree/importApi.ts
```diff
@@ -0,0 +1,496 @@
...
+    this.logger = Logger.childFromRoot(this.constructor.name);
+
+    this.sobjectUrlMap = new Map<string, string>();
+    this.schemaValidator = new SchemaValidator(this.logger, importPlanSchemaFile);
+  }
+
+  /**
+   * Inserts given SObject Tree content into given target Org.
+   *
+   * Validation and options fix-up done in execute() instead
```
WillieRuemmele, line 88: is this referencing the old `execute` methods that were on commands if so it should be `run` now right?

packages/plugin-data/src/api/data/tree/importApi.ts
```diff
@@ -0,0 +1,496 @@
...
+          msg = JSON.parse(err.message);
+          if (msg.hasErrors && msg.results && msg.results.length > 0) {
+            importResults.errors = msg.results;
+          }
+        } catch (e2) {
+          // throw original
+        }
+      }
+
+      throw err;
```
WillieRuemmele, line 179: should be `SfdxError.wrap(err)`

packages/plugin-data/src/api/data/tree/importApi.ts
```diff
@@ -0,0 +1,496 @@
...
+      } catch (err) {
+        if (err.name === 'ValidationSchemaFieldErrors') {
+          const e = SfdxError.create('@salesforce/plugin-data', 'importApi', 'dataPlanValidationError', [
+            planPath,
+            err.message,
+          ]);
+          e.name = INVALID_DATA_IMPORT_ERR_NAME;
+          throw e;
+        }
+        throw err;
```
WillieRuemmele, line 235: same SfdxError.wrap(err)

packages/plugin-data/src/api/data/tree/importApi.ts
```diff
@@ -0,0 +1,496 @@
...
+    // replace with reference found in previously saved records
+    if (resolveRefs && refMap) {
+      // find and stash all '@' references
+
+      // Could optionally use String.match(regex) to avoid turning off the rule
+      /* eslint-disable no-cond-assign */
+      while ((match = refRegex.exec(contentStr))) {
+        foundRefs.add(match[1]);
+      }
+      /* eslint-enable no-cond-assign */
```
WillieRuemmele, line 378: do we need this rule too?

packages/plugin-data/src/commands/force/data/tree/export.ts
```diff
@@ -0,0 +1,51 @@
...
+      description: messages.getMessage('prefix'),
+    }),
+    outputdir: flags.directory({
+      char: 'd',
+      description: messages.getMessage('outputdir'),
+    }),
+  };
+
+  // Overrides SfdxCommand.  This is ensured since requiresUsername == true
+  protected org!: Org;
```
WillieRuemmele, line 44: do we need to define this again? I thought it was brought in with SfdxCommad

packages/plugin-data/src/commands/force/data/tree/import.ts
```diff
@@ -0,0 +1,101 @@
...
+  id: string;
+}
+
+/**
+ * Command that provides data import capability via the SObject Tree Save API.
+ */
+export default class Import extends SfdxCommand {
+  public static readonly description = messages.getMessage('description');
+  public static readonly examples = messages.getMessage('examples').split(os.EOL);
+  public static readonly requiresProject = false;
```
WillieRuemmele, line 30: same as other command comment

packages/plugin-data/src/commands/force/data/tree/export.ts
shetzel: Unfortunately yes I think.  SfdxCommand.org is optional and to satisfy the TS compiler I had to override and tell it that `this.org` was not possibly undefined.  Let me know if you have another trick to handle this.

packages/plugin-data/src/commands/force/data/tree/export.ts
shetzel: You're right - thanks!  I'll change it.

packages/plugin-data/src/commands/force/data/tree/import.ts
shetzel: Yep!  thanks!

packages/plugin-data/src/api/data/tree/exportApi.ts
shetzel: They are.  this is shorthand TS syntactic sugar that will add them as readyonly instance vars.

packages/plugin-data/src/api/data/tree/exportApi.ts
shetzel: Because I didn't want to spend too much time typing everything when there are stories to handle refactoring all this into `@salesforce/data`, and I think this is just internal.  Still useful though so I'll change it.

packages/plugin-data/src/api/data/tree/exportApi.ts
shetzel: You're right, and I tried to but it was causing lots of issues so I punted.  There are more work items to handle things like this so unless we really need to I'd prefer to defer.

packages/plugin-data/src/api/data/tree/importApi.ts
shetzel: Good eye - thanks!

packages/plugin-data/src/commands/force/data/tree/export.ts
amphro: This is the trick I have been using too.

packages/plugin-data/src/api/data/tree/importApi.ts
shetzel: Probably not.  I'll try removing - thanks!

packages/plugin-data/src/api/data/tree/exportApi.ts
shetzel: because of linting.  There is a rule about a string template using a non-string.  The type of metadata.name is any, so I can either do `${metadata.name as string}` or the code above (or better typing overall, but that's another effort).

packages/plugin-data/src/api/data/tree/exportApi.ts
shetzel: I agree, and I tried but it causes a huge snowball that I thought best to handle in the refactor work items.

packages/plugin-data/src/api/data/tree/importApi.ts
shetzel: Oh good catch.  I renamed some methods within this class.  There used to be one called `execute`.  This comment should just go away because it refers to very legacy things.

packages/plugin-data/package.json
```diff
@@ -28,7 +28,15 @@
                 "description": "fetch records using SOQL"
               },
               "tree": {
-                "description": "manipulate records using the tree API"
+                "description": "import/export records using the tree API",
+                "subtopics": {
```
amphro, line 32: Why are these here? They aren't topics right?

packages/plugin-data/src/api/data/tree/exportApi.ts
shetzel: Yes but I'd like to save that for the refactor work item since it's not super straight forward.

amphro (approved):
I basically didn't really review this since I think you are doing as much of a copy/paste over as possible.

packages/plugin-data/package.json
shetzel: No they aren't topics.  It's a mistake.  I'll remove, thanks!

WillieRuemmele (approved):

PR #16, opened 2020-12-17
feat: @W-8449669@ migrate force:data:soql:query

peternhale:
### What does this PR do?
Migrattion of command force:data:soql:query to @salesforce/plugin-data
@W-8449669@


packages/plugin-data/package.json
```diff
@@ -70,6 +70,11 @@
     "@salesforce/core": "^2.13.0",
     "@salesforce/data": "^0.1.1",
     "@salesforce/ts-types": "^1.4.3",
+    "@types/chai-as-promised": "^7.1.3",
```
mshanemc, line 73: test-related stuff should be in devDependencies?

packages/plugin-data/src/api/data/soql/soqlQuery.ts
```diff
@@ -0,0 +1,82 @@
...
+    private readonly query: string,
+    private readonly logger: Logger
+  ) {
+    this.logger = logger.child('soqlQuery');
+  }
+
+  public async runSoqlQuery(): Promise<SoqlQueryResult> {
+    let columns: Field[] = [];
+    this.logger.debug('running query');
+    // the new way - autofetch, but not working right now for some reason
```
mshanemc, line 39: is it good now?

peternhale: @mshanemc yes it is. Removed commented code. Thank you!

packages/plugin-data/src/api/data/soql/soqlQuery.ts
```diff
@@ -0,0 +1,82 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { QueryResult } from 'jsforce';
+import { Connection, Logger } from '@salesforce/core';
+import { Tooling } from '@salesforce/core/lib/connection';
+import { Field, retrieveColumns } from './queryFields';
+
+/**
+ * Type to define SoqlQuert results
```
mshanemc, line 14: query
[thanks, https://github.com/streetsidesoftware/vscode-spell-checker]

peternhale: 👍 , but I use WebStorm, and it caught the typo. I chose to ignore it ;)

packages/plugin-data/src/api/data/soql/queryFields.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Class to represent a field when describing the fields that make up a query result
+ */
+export class Field {
+  public name: string;
+
+  public constructor(name: string) {
+    this.name = name;
+  }
+}
```
shetzel, line 20: Using a class for this doesn't feel quite right.  I think all of this will depend on what we want a public API to be.  Given this, I'd prefer to be conservative with what is exported so people don't try to use it directly.

packages/plugin-data/src/api/data/soql/queryFields.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Utility to fetch the columns involved in a soql query.
+ *
+ * Columns are then transformed into one of three types, Field, SubqueryField and FunctionField. List of
+ * fields is returned as the product.
+ *
+ * @param connection
+ * @param query
+ */
+export const retrieveColumns = async (connection: Connection | Tooling, query: string): Promise<Field[]> => {
```
shetzel, line 44: This seems like it's specific to command output right?  Do you expect people to want this outside of a command context?

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,140 @@
...
+import { CsvReporter, FormatTypes, HumanReporter, JsonReporter } from '../../../../api/data/soql/reporters';
+import { DataSoqlQueryResult } from '../../../../api/data/soql/dataSoqlQueryTypes';
+import { SoqlQuery, SoqlQueryResult } from '../../../../api/data/soql/soqlQuery';
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/plugin-data', 'soql.query');
+
+export class DataSoqlQueryCommand extends SfdxCommand {
+  public static readonly description = messages.getMessage('description');
+  public static readonly longDescription = messages.getMessage('longDescription');
```
shetzel, line 26: We no longer need a longDescription

packages/plugin-data/messages/soql.query.json
```diff
@@ -0,0 +1,22 @@
...
+  "description": "execute a SOQL query",
+  "longDescription": "Executes a SOQL query.",
+  "noResults": "Your query returned no results.",
+  "queryToExecute": "SOQL query to execute",
+  "queryLongDescription": "SOQL query to execute.",
+  "queryToolingDescription": "execute query with Tooling API",
+  "queryToolingLongDescription": "Execute the query using Tooling API.",
+  "queryInvalidReporter": "Unknown result format type. Must be one of the following values: %s",
+  "resultFormatDescription": "result format emitted to stdout; --json flag overrides this parameter",
+  "resultFormatLongDescription": "Format to use when displaying results. If you also specify the --json flag, --json overrides this parameter.",
```
shetzel, line 11: We no longer need any long descriptions

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,140 @@
...
+      default: false,
+      description: messages.getMessage('queryToolingDescription'),
+      longDescription: messages.getMessage('queryToolingLongDescription'),
+    }),
+    resultformat: flags.enum({
+      char: 'r',
+      description: messages.getMessage('resultFormatDescription'),
+      longDescription: messages.getMessage('resultFormatLongDescription'),
+      options: ['human', 'csv', 'json'],
+      default: 'human',
```
shetzel, line 52: We don't need `longDescription`, `required: false`, `hidden: false`, `default: false`

packages/plugin-data/src/dataCommand.ts
```diff
@@ -26,6 +26,9 @@ interface Result {
 // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
 // @ts-ignore because jsforce doesn't export http-api
 import * as HttpApi from 'jsforce/lib/http-api';
+import * as chalk from 'chalk';
+
+export const salesforceBlue = chalk.rgb(0, 161, 224);
```
shetzel, line 31: Are these changes necessary?  The query command doesn't extend this command so it appears to be unused.

packages/plugin-data/test/api/data/soql/queryFields.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import sinon = require('sinon');
+import { retrieveColumns } from '../../../../src/api/data/soql/queryFields';
+import * as TestUtil from '../../../testUtil';
+import { queryFieldsExemplars } from './test-files/queryFields.exemplars';
+
+chai.use(chaiAsPromised);
+
+describe('queryFields tests', () => {
+  // let toolingSpy: sinon.SinonSpy;
+  // let querySpy: sinon.SinonSpy;
```
shetzel, line 22: These can be removed.

packages/plugin-data/test/api/data/soql/queryFields.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  const fakeConnection = TestUtil.createFakeConnection();
+  let sandbox: any;
+  beforeEach(() => {
+    sandbox = sinon.createSandbox();
```
shetzel, line 26: sandboxes only need to be created once.  The sandbox state is restored in afterEach.

packages/plugin-data/test/api/data/soql/queryFields.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('should handle a query with just fields', async () => {
+    sandbox
+      .stub(fakeConnection, 'request')
+      .callsFake(() => Promise.resolve({ columnMetadata: queryFieldsExemplars.simpleQuery.columnMetadata }));
```
shetzel, line 35: This is more simply written as `.resolves({ columnMetadata: queryFieldsExemplars.simpleQuery.columnMetadata })`

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,140 @@
...
+      return {
+        data: toJsonMap(results.result),
+      };
+    }
+    return results;
+  }
+
+  private getConnection(org: Org, useToolingApi: boolean): Connection | Tooling {
+    return useToolingApi ? org.getConnection().tooling : org.getConnection();
+  }
```
shetzel, line 139: Is this a method for easier testing?  With it being just 1 line, seems like it could just go within `run()` and maybe you wouldn't need to import `Tooling` in that case?

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,140 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+// This is the legacy converted command file. Ignoring code-coverage since this is generated.
+// THIS SHOULD BE REMOVED WHEN CONVERTED TO EXTEND SfdxCommand
+/* istanbul ignore file */
```
shetzel, line 10: We can delete this comment now

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,140 @@
...
+   * When the user is using global '--json' flag an instance of SfdxResult si returned.
+   * This is necessary since '--json' flag reports results in the form of SfdxResult
+   * and bypasses the definition of start result. The goal is to have the output
+   * from '--json' and '--resulformat json' be the same.
+   *
+   * The DataSoqlQueryResult is necessary to communicate user selections to the reporters.
+   * The 'this' object available during display() function does not include user input to
+   * the command, which are necessary for reporter selection.
+   *
+   */
```
shetzel, line 99: Maybe we should modify SfdxResult to support the usecase so we don't have to modify the return of `run`?  I think it was just an oversight on my part when I wrote that long ago.

shetzel (changes requested):

packages/plugin-data/test/testUtil.ts
```diff
@@ -0,0 +1,342 @@
...
+    rawListeners(event: string | symbol) {
+      return [];
+    },
+    eventNames(): Array<string | symbol> {
+      return [];
+    },
+  };
+};
+
+export const createBaseFakeConnection = function (): BaseConnection {
```
WillieRuemmele, line 78: do we really need all of these methods stubbed? same with `Emitter` above. and `Connection` below.

packages/plugin-data/package.json
```diff
@@ -65,21 +65,22 @@
     "test": "sf-test"
   },
   "dependencies": {
-    "@oclif/config": "^1",
     "@salesforce/command": "^3.0.3",
     "@salesforce/core": "^2.13.0",
-    "@salesforce/data": "^0.1.1",
     "@salesforce/ts-types": "^1.4.3",
+    "chalk": "^4.1.0",
```
WillieRuemmele, line 71: What is chalk being used for? I think it can be accessed via `this.ux` in the command

packages/plugin-data/src/api/data/soql/reporters.ts
```diff
@@ -0,0 +1,326 @@
...
+      });
+    } else {
+      this.logger.info(`No fields found for query "${this.data.query}"`);
+    }
+
+    return { attributeNames, children, aggregates };
+  }
+
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  public soqlQuery(columns: string[], records: any[], totalCount: number): void {
```
WillieRuemmele, line 104: can we create a `record` type? it seems it's passed as any[] into most functions - just nice to add types where we can

packages/plugin-data/src/api/data/soql/reporters.ts
```diff
@@ -0,0 +1,326 @@
...
+    this.prepNullValues(records);
+    this.ux.table(records, { columns: this.prepColumns(columns) });
+    this.log(chalk.bold(messages.getMessage('displayQueryRecordsRetrieved', [totalCount])));
+  }
+
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  public prepNullValues(records: any[]): void {
+    records.forEach((record): void => {
+      Reflect.ownKeys(record).forEach((propertyKey) => {
```
WillieRuemmele, line 113: Where is `Reflect` coming from? If it is a library can we get workaround using it?

packages/plugin-data/src/api/data/soql/reporters.ts
```diff
@@ -0,0 +1,326 @@
...
+            }
+          });
+        }
+        return newResults;
+      }, []);
+    }
+    return queryResults;
+  }
+}
+// /* eslint-enable @typescript-eslint/no-explicit-any */
```
WillieRuemmele, line 177: what any is this rule ignoring? I _think_ it might be extra?

packages/plugin-data/src/commands/force/data/soql/query.ts
peternhale: @shetzel, I agree, but should that be handled in a different user story?

packages/plugin-data/test/api/data/soql/reporter.test.ts
```diff
@@ -0,0 +1,172 @@
...
+import { expect } from '@salesforce/command/lib/test';
+import * as chai from 'chai';
+import * as chaiAsPromised from 'chai-as-promised';
+import sinon = require('sinon');
+import { UX } from '@salesforce/command';
+import { Logger } from '@salesforce/core';
+import { SinonSandbox } from 'sinon';
+import { soqlQueryExemplars } from '../../../api/data/soql/test-files/soqlQuery.exemplars';
+import { DataSoqlQueryResult } from '../../../../src/api/data/soql/dataSoqlQueryTypes';
+import { HumanReporter } from '../../../../src/api/data/soql/reporters';
```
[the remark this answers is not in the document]
peternhale: @WillieRuemmele, three different sources. Are you suggesting that the sources be combined or am I missing a TS/JS trick here?

packages/plugin-data/src/api/data/soql/reporters.ts
peternhale: @WillieRuemmele, I will take a look at it.

packages/plugin-data/src/api/data/soql/reporters.ts
peternhale: @WillieRuemmele [Reflect](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect) is part of JS and I prefer it over using Objects. What I would look for to replace it would be equivalent functions from our own libraries.

I will look for suitable replacements.

packages/plugin-data/src/api/data/soql/reporters.ts
peternhale: @WillieRuemmele, extra it is. Thanks!

packages/plugin-data/package.json
peternhale: @WillieRuemmele, chalk is being used to apply display attributes to text that is written to the console. Display attributes (color, bolding, etc) in toolbelt was accomplished by accessing the color property in the available logger instance. That property does not seem to be present on the available Logger class from @salesforce/core.

Open to suggestions and edification :).

packages/plugin-data/src/commands/force/data/soql/query.ts
peternhale: I will inline the ternary and remove the private function.

packages/plugin-data/src/api/data/soql/reporters.ts
mshanemc: `Reflect` is new to me, but it looks handy.

packages/plugin-data/src/commands/force/data/soql/query.ts
shetzel: Yeah, that's a good idea.

packages/plugin-data/package.json
shetzel: chalk was removed from the `UX` class a while ago on purpose.  Not everyone needs styling with their terminal output and it's easy to add the dependency to use it.  This is the expected approach for adding styling.

packages/plugin-data/src/api/data/soql/reporters.ts
shetzel: This is approaching personal preference so not sure it's worth spending much time discussing it, but we might want to bring it up as a team during planning or a parking lot, at least so that everyone on the team is aware.  There's a balance between code consistency for easier maintenance and freedom to write code to your style.  This is really only relevant when object properties are symbols or non-enumerable.  Maybe we should add something to the `kit` library for this if there isn't something there already?

peternhale: @shetzel, I am all for providing this functionality in kit.

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,224 @@
...
+  };
+
+  // Overrides SfdxCommand.  This is ensured since requiresUsername == true
+  protected org!: Org;
+
+  /**
+   * Command run implementation
+   *
+   * Returns either a DataSoqlQueryResult or a SfdxResult.
+   * When the user is using global '--json' flag an instance of SfdxResult si returned.
```
amphro, line 178: `si` -> `is`

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * This function maps the DataSoqlQueryResult instance to the the well known 'data' property in
+   * SfdxResult class when the user has selected '--json' flag.
+   *
+   * @param results
+   * @private
+   */
+  private normalizeIfJson(results: DataSoqlQueryResult): DataSoqlQueryResult | SfdxResult {
```
amphro, line 216: This whole thing seems wrong to me. You shouldn't need to do anything with the `data` property. What were you running into?

image

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,224 @@
...
+      if (this.flags.resultformat !== 'json') this.ux.startSpinner(messages.getMessage('queryRunningMessage'));
+      const query = new SoqlQuery();
+      const queryResult: SoqlQueryResult = await query.runSoqlQuery(
+        this.flags.usetoolingapi ? this.org.getConnection().tooling : this.org.getConnection(),
+        this.flags.query,
+        this.logger
+      );
+      const results = {
+        ...queryResult,
+        resultFormat: this.flags.resultformat,
```
amphro, line 199: I don't see any of these in the old json output. Did you just decide to add them?

packages/plugin-data/test/queryFields.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import sinon = require('sinon');
+import { SoqlQuery } from '../lib/commands/force/data/soql/query';
+import * as TestUtil from './testUtil';
+import { queryFieldsExemplars } from './test-files/queryFields.exemplars';
+
+chai.use(chaiAsPromised);
+
+describe('queryFields tests', () => {
+  // let toolingSpy: sinon.SinonSpy;
```
amphro, line 21: remove?

packages/plugin-data/test/soqlQuery.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('should handle a simple query with all records returned in single call', async () => {
+    sandbox
+      .stub(fakeConnection, 'request')
+      .resolves({ columnMetadata: queryFieldsExemplars.simpleQuery.columnMetadata });
+    querySpy = sandbox.stub(fakeConnection, 'autoFetchQuery').resolves(soqlQueryExemplars.simpleQuery.queryResult);
+    // queryMoreSpy = sandbox.stub(fakeConnection, 'queryMore');
```
amphro, line 36: remove all these type of comments in the tests?

packages/plugin-data/test/testUtil.ts
amphro: We should be able to switch this to `stubInterface` or `stubObject` from `@salesforce/ts-sinon`.

amphro (changes requested):

packages/plugin-data/src/commands/force/data/soql/query.ts
peternhale: @amphro The current version of the query command emitted user output during the execution of the command. I moved output to the override of the `display()` function. Currently the `this` reference made available to `display()` does not include the command parameters, etc. made by the user. I am passing along the needed data to drive selected output format.

packages/plugin-data/src/commands/force/data/soql/query.ts
peternhale: @amphro , This ties into the additional properties I am passing in the result. There are two ways to ask for json results; `resultformat json` and `--json`. When --json is used the display() call is skipped by command or oclif (not sure which). Given that I have added additional properties to the result, this function "restores" the expected result structure when `--json` is given. Doing this ensures the json displayed using resultformat json is exactly the same as when using --json.

In short, varying the output by user selection is difficult to do using `display()` if the data to help shape the output is not available in the standard `this` object.

packages/plugin-data/test/testUtil.ts
peternhale: I will look into this, thanks.

packages/plugin-data/src/commands/force/data/soql/query.ts
amphro: Ya, I just don't know if this is the best way to do it, because this becomes part of the output of `--json` right? For example, adding logger on the result will put the whole logger object in the json that is output from the terminal? Seems like we don't want that.

amphro: If it doesn't output that information, then I guess it is fine. It seems like we are trying to fit a square peg in a round hole with the `display` function though. Maybe it would have been better just to call `reporter.display` in the run method instead...

amphro: Or, I think you can also just say 

```
this.result = result;
return queryResult;
```

rather than having to have the normalizeJson method. This also prevents having to have two return types from the run method which will be confusing when tooling vendors consume these types which should alway mirror what is returned from the run method.

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,198 @@
...
+   * This is necessary since '--json' flag reports results in the form of SfdxResult
+   * and bypasses the definition of start result. The goal is to have the output
+   * from '--json' and '--resulformat json' be the same.
+   *
+   * The DataSoqlQueryResult is necessary to communicate user selections to the reporters.
+   * The 'this' object available during display() function does not include user input to
+   * the command, which are necessary for reporter selection.
+   *
+   */
+  public async run(): Promise<SfdxResult> {
```
amphro, line 160: We do not want to return `SfdxResults`. If you look at the existing `--json` output, there is not a `data` property. It should be `QueryResult<Record>` from jsforce.

shetzel (approved):

packages/plugin-data/src/dataCommand.ts
peternhale: I'll revert the change

mshanemc:
QA findings:
`select count()` didn't work on the old plugin (returns 0 records even when they exist).  On this command, it has a node warning:

```

**Comment by mshanemc** [comment]
QA notes:
the old plugin didn't handle `select count() from` at all.  It'd return 0 records but act like everything was good.
This one returns a confusing node warning (this org has 219 of these object records).
```
$ bin/run force:data:soql:query -u platformers -q "select count() from Demo__c"
(node:6653) Warning: The query result is missing 219 records due to a 50000 record limit. Increase the number of records returned by setting the config value "maxQueryLimit" or the environment variable "SFDX_MAX_QUERY_LIMIT" to 219 or greater than 50000.


Total number of records retrieved: 0.
Querying Data... done
```

**Comment by mshanemc** [comment]
QA notes:
The json output of the command changed...here's a screenshot side-by-side of the before/after result of a query with --json.

I like that the library is giving back the query and the column structure...but we'll need to parse that in the command to keep the json consistent with toolbelt's json.

Screen Shot 2021-01-21 at 2 26 50 PM

**Comment by peternhale** [comment]
@mshanemc I think the difference was introduced by the move from query/with queryMore to autoFetch. In order to get to parity with old version, the maxFetch is always set to 50000. I believe this message is being emitted from `autoFetchQuery` logic. I see there was an issue reported [here](https://github.com/forcedotcom/cli/issues/554) that is related.

**Comment by mshanemc** [comment]
QA notes:
Here's a screenshot of the json output of a soql w/ subquery (you can kinda see the example in the new query property).

1. the new command is omitting the child records
2. note that they also have the old jsforce data structure (done, totalSize, records) **on the child item object** 

Screen Shot 2021-01-21 at 2 33 34 PM

**Comment by peternhale** [comment]
> QA notes:
> The json output of the command changed...here's a screenshot side-by-side of the before/after result of a query with --json.
> 
> I like that the library is giving back the query and the column structure...but we'll need to parse that in the command to keep the json consistent with toolbelt's json.
> 
> Screen Shot 2021-01-21 at 2 26 50 PM

@mshanemc, I should remove properties `query` and `columns` from that return. The property `data` comes from SfdxResult and whatever the run method returns is assigned to that property.

```
export declare class Result implements SfdxResult {
    data: AnyJson;
    tableColumnData?: TableOptions;
    ux: UX;
    constructor(config?: SfdxResult);
    display(): void;
}
```

**Comment by peternhale** [comment]
> QA notes:
> Here's a screenshot of the json output of a soql w/ subquery (you can kinda see the example in the new query property).
> 
> 1. the new command is omitting the child records (interestingly, subquery records do correctly appear in the non-json output)
> 2. note that they also have the old jsforce data structure (done, totalSize, records) **on the child item object**
> 
> Screen Shot 2021-01-21 at 2 33 34 PM

I will look into this delta

**Comment by mshanemc** [comment]
QA notes:
There's also a different in the json output of a groupby (aggregate) query...the new version has the expr(0) but not the count(id).  The above-mentioned differences are still there, so I'm now just focusing on the records:[] portion. 
 
The query I used to test is `select count(id), Demo_Type__c from Demo__c group by Demo_Type__c`
Screen Shot 2021-01-21 at 2 48 09 PM

**Comment by mshanemc** [comment]
QA notes:
when outputting --json to a file, the old command saved null values as null and new one saves them as a string with value "null"
The version number also changed, but that may be the result of running `sfdx` vs. `bin/run` for comparison.

`select LastLoginDate,Namespace,Id,Description from ScratchOrgInfo limit 10`

Screen Shot 2021-01-21 at 2 55 17 PM

**Comment by mshanemc** [comment]
QA notes:

in the terminal output of `force:data:soql:query -u prehub -q "select LastLoginDate,Namespace,Id,Description from ScratchOrgInfo limit 10" --json` the new command (left) is getting ansi-color-codes mixed into the nulls

Screen Shot 2021-01-21 at 3 01 02 PM

**Comment by peternhale** [comment]
> QA notes:
> There's also a difference in the json output of a groupby (aggregate) query...the new version has the expr(0) but not the count(id). The above-mentioned differences are still there, so I'm now just focusing on the records:[] portion.
> 
> The query I used to test is `select count(id), Demo_Type__c from Demo__c group by Demo_Type__c`
> Screen Shot 2021-01-21 at 2 48 09 PM

Will look into the delta

**Comment by peternhale** [comment]
> QA notes:
> 
> in the terminal output of `force:data:soql:query -u prehub -q "select LastLoginDate,Namespace,Id,Description from ScratchOrgInfo limit 10" --json` the new command (left, sorry it's backwards from the other screenshots) is getting ansi-color-codes mixed into the nulls
> 
> Screen Shot 2021-01-21 at 3 01 02 PM

This was caused by same issue that was stripping subquery records from the result.

**Comment by mshanemc** [comment]
> > QA notes:
> > The json output of the command changed...here's a screenshot side-by-side of the before/after result of a query with --json.
> > I like that the library is giving back the query and the column structure...but we'll need to parse that in the command to keep the json consistent with toolbelt's json.
> > Screen Shot 2021-01-21 at 2 26 50 PM
> 
> @mshanemc, I should remove properties `query` and `columns` from that return. The property `data` comes from SfdxResult and whatever the run method returns is assigned to that property.
> 
> ```
> export declare class Result implements SfdxResult {
>     data: AnyJson;
>     tableColumnData?: TableOptions;
>     ux: UX;
>     constructor(config?: SfdxResult);
>     display(): void;
> }
> ```
> 
> Additional properties removed. now conforms to old results.

Is there a reason we should use SfdxResult @amphro ?  I usually just return the results object (at the very end) and do table formatting/output with ux.table, and that's what I've seen in other plugins (auth, user)

**Comment by peternhale** [comment]
> > > QA notes:
> > > The json output of the command changed...here's a screenshot side-by-side of the before/after result of a query with --json.
> > > I like that the library is giving back the query and the column structure...but we'll need to parse that in the command to keep the json consistent with toolbelt's json.
> > > Screen Shot 2021-01-21 at 2 26 50 PM
> > 
> > 
> > @mshanemc, I should remove properties `query` and `columns` from that return. The property `data` comes from SfdxResult and whatever the run method returns is assigned to that property.
> > ```
> > export declare class Result implements SfdxResult {
> >     data: AnyJson;
> >     tableColumnData?: TableOptions;
> >     ux: UX;
> >     constructor(config?: SfdxResult);
> >     display(): void;
> > }
> > ```
> > 
> > 
> > Additional properties removed. now conforms to old results.
> 
> Is there a reason we should use SfdxResult @amphro ? I usually just return the results object (at the very end) and do table formatting/output with ux.table, and that's what I've seen in other plugins (auth, user)

@mshanemc The error seen was my error. I originally had Promise<SomeOtherType|SfdxResult> for the return. Thomas encouraged me not to do it this way. I removed the SomeOtherType but did not exclude the unique properties found in SomeOtherType, which you discovered.

File: packages/plugin-data/src/commands/force/data/soql/query.ts [hunk 28]
```diff
@@ -0,0 +1,201 @@
...
+ * Class to handle a soql query
+ *
+ * Will collect all records and the column metadata of the query
+ */
+export class SoqlQuery {
+  public async runSoqlQuery(connection: Connection | Tooling, query: string, logger: Logger): Promise<SoqlQueryResult> {
+    let columns: Field[] = [];
+    logger.debug('running query');
+
+    const result = await connection.autoFetchQuery(query, { autoFetch: true, maxFetch: 50000 });
```
peternhale, line 36: @amphro, I am using the default maxFetch of 50000. This is the only way I could get on par with old query command that used query/queryMore up to 5000 records.

I learned today that there is a config setting `maxQueryLimit`. Should I be considering that setting and if not present, default to 50000 (or maybe 10000, which is a published default)?

amphro: Can we just call `autoFetchQuery` with no options and leave everything to be the default? That warning message about the records returned is possibly a bug in core around `count()`. FYI @WillieRuemmele might be able to provide more info.

peternhale: @amphro, the default for maxFetch is 10000 records and maxQueryLimit is considered in autoFetch impl, however original behavior was query/queryMore until records were exhausted, delegating how many rows were returned to the query itself, i.e. `select id from Account LIMIT 1` or `select id from Account`, or `select id from Account LIMIT 49999`. Given 50000 records is the hard limit for a soql query, it seemed like the reasonable thing to do, while still allowing user to control records returned with `LIMIT` in queries.

So, removing maxFetch from soql query may change existing behavior. We would need to let customers know that in a release note.

packages/plugin-data/src/commands/force/data/soql/query.ts
mshanemc: @peternhale let's drop `default:false` for booleans

packages/plugin-data/src/commands/force/data/soql/query.ts
amphro: I see. I didn't understand how it worked before then. I would prefer to keep the behavior how it was before unless we have a good reason (performance, security, etc).

packages/plugin-data/src/commands/force/data/soql/query.ts
```diff
@@ -0,0 +1,187 @@
...
+   * This is necessary since '--json' flag reports results in the form of SfdxResult
+   * and bypasses the definition of start result. The goal is to have the output
+   * from '--json' and '--resulformat json' be the same.
+   *
+   * The DataSoqlQueryResult is necessary to communicate user selections to the reporters.
+   * The 'this' object available during display() function does not include user input to
+   * the command, which are necessary for reporter selection.
+   *
+   */
+  public async run(): Promise<unknown> {
```
amphro, line 147: You don't need to return `unknown`. You just can't return `SfdxResult`. For example, returning `Promise<QueryResult<Record>>` would work just fine.

amphro (approved):

peternhale:
@amphro I had to return unknown in order to keep the JSON results the same as toolbelt. SfdxResult adds in `data` property. If you have a look at the other data commands, they all return `unknown`.

packages/plugin-data/src/commands/force/data/soql/query.ts
peternhale: @amphro It is still in its original form,`maxFetch: 50000`

amphro:
> @amphro I had to return unknown in order to keep the JSON results the same as toolbelt. SfdxResult adds in `data` property. If you have a look at the other data commands, they all return `unknown`.
> 
> [#16 (comment)](https://github.com/salesforcecli/data/pull/16#discussion_r564216194)

I'm saying do NOT use SfdxResult, but you can use other return types, like `Promise<QueryResult<Record>>` which works fine.

peternhale:
> > @amphro I had to return unknown in order to keep the JSON results the same as toolbelt. SfdxResult adds in `data` property. If you have a look at the other data commands, they all return `unknown`.
> > [#16 (comment)](https://github.com/salesforcecli/data/pull/16#discussion_r564216194)
> 
> I'm saying do NOT use SfdxResult, but you can use other return types, like `Promise<QueryResult<Record>>` which works fine.

Ok if I leave it as is, matches the other data commands? ;).

PR #23, opened 2021-01-30
Update exportApi.ts

robert-strunk:
### What does this PR do?
Removes the string mutation that is changing the case of any IDs that may be in a query filter. 
It then uses regex to see if the query starts with "select" (ignoring case). 

Should resolve bug related to W-8827929 - https://github.com/forcedotcom/cli/issues/828


amphro (changes requested):
Thanks for the PR. Looks good. Can you add a unit test for a query with ID? Then we can test and merge this in.

robert-strunk:
@amphro this is the first time I've tried to test Javascript. Please review this accordingly. I'm always looking to improve so any feedback at all will be happily accepted.

packages/plugin-data/src/api/data/tree/exportApi.ts
```diff
@@ -115,8 +115,10 @@ export class ExportApi {
       }
     }
 
-    config.query = config.query.toLowerCase().trim();
-    if (!config.query.startsWith('select')) {
+    config.query = config.query.trim();
+    let startsWithSelectPattern = new RegExp('^select', 'i');
+
+    if (!config.query.match(startsWithSelectPattern)) {
```
peternhale, line 121: This could use the `/regex/.test('foo')` form of regex to accomplish the same and eliminates the need for `startsWithSelectPattern` variable.

`if (!/^select/i.test(config.query)) ...`

peternhale (changes requested):

packages/plugin-data/src/api/data/tree/exportApi.ts
robert-strunk: Changes made.

peternhale:
@robertStrunk Appears the check failures are lint errors.

From the repository root directory run `yarn test` to see the errors locally.

robert-strunk:
@peternhale  Thanks! I have worked through the errors with the exception of one. In the `data\packages\plugin-data\test\queryFields.test.ts` file it is trying to import a module from a non-existent directory `data\packages\plugin-data\lib\commands\force\data\soql\query`. since this is existing code I feel like there was some local command I failed to run to generate that directory. Any ideas?

image

peternhale:
@robert-strunk, I tried this branch locally and did not see the error regarding a missing file.

Try running `yarn run prepack` and then `yarn run test` to see if that resolves the issue.

robert-strunk:
That command just gives me a slew of errors. Unfortunately I don't have time to figure all this out so delete this pull request if you like.

peternhale:
@robert-strunk, I took your changes and cleaned up the issues. PR #35 is up for review.

Thank you for the contribution and I will be closing this PR.

robertStrunk:
Very much appreciated! New to this model of development, a bit overwhelmed with it all.

peternhale:
@robert-strunk when you have a free moment it would be nice to hear about your struggles in this new model of development. If possible list the tools that you are using and any documentation/readme, etc that you found useful or found not so useful.

PR #31, opened 2021-02-01
fix: parse csv as utf-8 and enable bom parsing

WillieRuemmele:
### What does this PR do?
passes csv to utf-8 encoding, and then adds the bom parsing option to the parser, which is now safe because the files are utf-8 encoded.

@W-8836961@
https://github.com/forcedotcom/cli/issues/720

### DIFF
--- a/packages/plugin-data/src/commands/force/data/bulk/delete.ts
+++ b/packages/plugin-data/src/commands/force/data/bulk/delete.ts
@@ -46,7 +46,7 @@ export default class Delete extends DataCommand {
       const conn: Connection = this.org.getConnection();
       this.ux.startSpinner('Bulk Delete');
 
-      const csvRecords: ReadStream = fs.createReadStream(this.flags.csvfile);
+      const csvRecords: ReadStream = fs.createReadStream(this.flags.csvfile, { encoding: 'utf-8' });
       const job: Job = conn.bulk.createJob(this.flags.sobjecttype, 'delete');
 
       const batcher: Batcher = new Batcher(conn, this.ux);
--- a/packages/plugin-data/src/batcher.ts
+++ b/packages/plugin-data/src/batcher.ts
@@ -303,6 +303,7 @@ export class Batcher {
         // library option is snakecase
         // eslint-disable-next-line @typescript-eslint/camelcase
         skip_empty_lines: true,
+        bom: true,
       });
 
       readStream.pipe(parser);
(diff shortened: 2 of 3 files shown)


mshanemc (approved):

packages/plugin-data/src/commands/force/data/bulk/delete.ts
```diff
@@ -46,7 +46,7 @@ export default class Delete extends DataCommand {
       const conn: Connection = this.org.getConnection();
       this.ux.startSpinner('Bulk Delete');
 
-      const csvRecords: ReadStream = fs.createReadStream(this.flags.csvfile);
+      const csvRecords: ReadStream = fs.createReadStream(this.flags.csvfile, { encoding: 'utf-8' });
```
peternhale, line 49: @WillieRuemmele, I know utf-8 is very common, but is it exclusive?

WillieRuemmele: I believe it is the standard for the CLI

peternhale: What I am concerned about is a user that creates a CSV file **_outside_** of cli with an encoding that is not utf-8. Chances are slim, but Murphy is always at the ready :).

shetzel (commented):
What does the bom option do?  Why does this require no unit test changes or additions?

tells the csv-parser to read the byte-order-mark from the csv if it exists.  Without it, csvs that include the BOM create problems because the BOM is considered part of the data.

packages/plugin-data/src/commands/force/data/bulk/delete.ts
mshanemc: The default for fs.readStream is `null` which means "don't try to deal with encoding at all."  Presumably, those other encodings fail today.

From testing, it **looks** like files with no encoding work fine with utf8.  

[for context, the csv that led to this (utf8 with byte order mark) is generated by Tableau's tooling]

WillieRuemmele:
@shetzel see https://github.com/forcedotcom/cli/issues/720 he posts a good breakdown of why/what it is. Sounds like it comes from the output of a tableau tool. From my understanding, setting `bom:true` in the parser options will strip it out if it's there ( which we want) otherwise it won't do anything

packages/plugin-data/src/commands/force/data/bulk/delete.ts
peternhale: Given this particular use case, should the commands be extended to expose file encoding, defaulting to utf-8? Maybe with a user story to, instead of this bug fix?

Is there any error telemetry data that shows the kind of errors encountered to see if adding an option might be required?

WillieRuemmele: we also officially say that csv files must be in utf-8 https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/datafiles_csv_preparing.htm

peternhale (approved):

shetzel (approved):

PR #35, opened 2021-02-04
fix: get roberts pr healthy

peternhale:
@W-8827929@

### What does this PR do?
export was causing making the query all lowercase, damaging case sensitive data


jwalke:
@amphro @peternhale Hi - as a customer I'd love to see this fix get merged in. We've run into it a few times over the years and I was just doing some digging today to see if it was something I could help fix but found that this PR already looks like it does exactly that. Thanks!

amphro (approved):

amphro:
The change looks good to me.

PR #40, opened 2021-02-22
fix: commands ignore --usetoolingapi

mshanemc:
### What does this PR do?
fix casing to honor the flag `usetoolingapi`

@W-8879900@

### DIFF
--- a/packages/plugin-data/src/dataCommand.ts
+++ b/packages/plugin-data/src/dataCommand.ts
@@ -94,7 +94,7 @@ export abstract class DataCommand extends SfdxCommand {
   }
 
   public getConnection(): BaseConnection {
-    const connection: BaseConnection & ConnectionInternals = this.flags.useToolingApi
+    const connection: BaseConnection & ConnectionInternals = this.flags.usetoolingapi
       ? this.org.getConnection().tooling
       : this.org.getConnection();
 


peternhale (approved):

packages/plugin-data/src/dataCommand.ts
```diff
@@ -94,7 +94,7 @@ export abstract class DataCommand extends SfdxCommand {
   }
 
   public getConnection(): BaseConnection {
-    const connection: BaseConnection & ConnectionInternals = this.flags.useToolingApi
+    const connection: BaseConnection & ConnectionInternals = this.flags.usetoolingapi
```
WillieRuemmele, line 97: do all commands that have that flag extend `dataCommand` and use this `getConnection` method?

peternhale: @WillieRuemmele create/update/delete/get/query use the flag usertoolingapi

mshanemc: yes, they do.

WillieRuemmele: yea, but a few weeks ago `soql:query` didn't extend `DataCommand` and used `this.org.getConnection` not `this.getConnection` I just want to make sure this code is being called for all commands that could use it

WillieRuemmele (approved):

PR #85, opened 2021-04-09
chore: fix length assertion

peternhale:
### What does this PR do?
Loosens the assert when checking the number of record from an exact match to an "above" match. This assertion fails at times when the async bulk delete reports as complete, but is really not. This is a server side issue, so accounting for the possibility that this may crop up in a test.
@W-0@


packages/plugin-data/package.json
```diff
@@ -69,7 +69,7 @@
     "test": "sf-test",
     "test:command-reference": "./bin/run commandreference:generate --erroronwarnings",
     "test:deprecation-policy": "./bin/run snapshot:compare",
-    "test:nuts": "nyc mocha \"**/*.nut.ts\" --slow 4500 --timeout 600000 --parallel",
+    "test:nuts": "nyc mocha \"**/*.nut.ts\" --slow 4500 --timeout 600000",
```
mshanemc, line 72: why?

mshanemc (commented):
why disable parallel?

RodEsp:
I second Shane's question ^.

peternhale:
> why disable parallel?
I'll revert

shetzel (approved):

RodEsp (approved):

PR #110, opened 2021-05-25
Sm/record get where boolean w 6236756

mshanemc:
### What does this PR do?
turns strings that appear to be booleans into booleans 
tests with a NUT

@W-6236756@


WillieRuemmele:
has this always been a bug? I saw the chatter thread was 2 years old... just surprised no one else has noticed this

WillieRuemmele (approved):

mshanemc:
> has this always been a bug? I saw the chatter thread was 2 years old... just surprised no one else has noticed this

yes.  I just don't think there's that many people using the command.

packages/plugin-data/src/dataCommand.ts
```diff
@@ -169,13 +169,19 @@ export abstract class DataCommand extends SfdxCommand {
...
-        constructedObject[key] = pair.substr(eqPosition + 1);
+        constructedObject[key] = this.convertToBooleanIfApplicable(pair.substr(eqPosition + 1));
       }
     });
 
     return constructedObject;
   }
 
+  private convertToBooleanIfApplicable(input: string): boolean | string {
+    if (input.toLowerCase() === 'false') return false;
```
peternhale, line 180: @mshanemc You may want to trim the input. Not sure how strict key/value pairs are treated on the command line. 
```
sfdx force:data:record:get -s Account -w "Name='Universal Containers' Foo=true"
sfdx force:data:record:get -s Account -w "Name='Universal Containers' Foo= true"
```

peternhale (changes requested):

packages/plugin-data/src/dataCommand.ts
mshanemc: good point.  I put it in `transformKeyValueSequence`

peternhale: @mshanemc Not sure that is the right place. What if there are intentional leading or trailing whitespace is string values?

peternhale (changes requested):

peternhale (approved):

PR #143, opened 2021-06-17
sm/ts-next-corrections

mshanemc:
### What does this PR do?
fixes TS/ES-next errors from overriding base sfdxCommand org property

@W-9329741@

### DIFF
--- a/packages/plugin-data/src/dataCommand.ts
+++ b/packages/plugin-data/src/dataCommand.ts
@@ -7,7 +7,7 @@
 
 import { SfdxCommand } from '@salesforce/command';
 import { AnyJson, Dictionary, get, Nullable } from '@salesforce/ts-types';
-import { fs, Messages, SfdxError } from '@salesforce/core';
+import { fs, Messages, SfdxError, Org } from '@salesforce/core';
 import { BaseConnection, ErrorResult, Record, SObject } from 'jsforce';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore because jsforce doesn't export http-api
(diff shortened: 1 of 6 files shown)


WillieRuemmele (commented):
will other plugins need to create their own `ensureOrg`? will the upgrades affect `SfdxCommand` in a similar way?

mshanemc:
> will other plugins need to create their own `ensureOrg`? will the upgrades affect `SfdxCommand` in a similar way?

You can normally just add `as Org` in your command if you know it's going to exist and you're merely extending SfdxCommand.

I think this one was weird because it was putting the "requiresUsername" stuff in a dataCommand class and then extending for the actual commands.  You can't override Org at that level of inheritance.

I could have moved all the requiresUsername stuff to all the commands as another option, but this seemed more in line with what Pete was writing (a fat dataCommand that does all the work).

packages/plugin-data/src/dataCommand.ts
```diff
@@ -96,10 +94,17 @@ export abstract class DataCommand extends SfdxCommand {
     return final;
   }
 
+  public ensureOrg(): Org {
+    if (!this.org) {
+      throw new Error('An org is required for a connection');
```
peternhale, line 99: @mshanemc I would prefer the error be the same that is emitted when an org cannot be found.

peternhale (changes requested):
@mshanemc what TS error is this addressing?

packages/plugin-data/src/dataCommand.ts
mshanemc: done

WillieRuemmele (approved):

PR #166, opened 2021-07-20
feat: added new flag for serial process

vamsimundra:
### What does this PR do?
Added a new boolean flag --serial for the "force:data:bulk:upsert" command,  which will allow the job to run serially. 
Github Issue Link : https://github.com/forcedotcom/cli/issues/882
@W-8934566@

### DIFF
--- a/packages/plugin-data/src/commands/force/data/bulk/upsert.ts
+++ b/packages/plugin-data/src/commands/force/data/bulk/upsert.ts
@@ -40,6 +40,11 @@ export default class Upsert extends DataCommand {
       description: messages.getMessage('flags.wait'),
       min: 0,
     }),
+    serial: flags.boolean({
+      char: 'r',
+      description: messages.getMessage('flags.serial'),
+      default: false,
+    }),
   };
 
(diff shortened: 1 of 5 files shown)


WillieRuemmele (commented):
Nice Job!

WillieRuemmele (approved):

packages/plugin-data/src/commands/force/data/bulk/upsert.ts
```diff
@@ -40,6 +40,11 @@ export default class Upsert extends DataCommand {
       description: messages.getMessage('flags.wait'),
       min: 0,
     }),
+    serial: flags.boolean({
+      char: 'r',
+      description: messages.getMessage('flags.serial'),
+      default: false,
```
mshanemc, line 46: you can leave off `default:false` since that's the automatic value for booleans when they aren't there.

mshanemc (approved):
very nice tests.

in the PR description, can you put a link to the github WI?  That makes this PR show up on the WI as a "mention" so anyone who's following that sees that it's being done.

PR #249, opened 2021-12-28
Wr/validation error

WillieRuemmele:
\#\#\# What does this PR do?
includes the error message from the server in the error message shown to the user

\#\#\# What issues does this PR fix or reference?
@W\-10322646@
https://github.com/forcedotcom/cli/issues/1327


mshanemc (approved):

mshanemc:
QA Notes:
✅ tested with multiple validation rules (only the first one shows up).
✅ fields is populated only when the error isn't top of page...otherwise it's an empty array []

PR #266, opened 2022-02-01
Wr/data bugs

WillieRuemmele:
### What does this PR do?
fixes various bugs with `data`
lots of the QA steps will be found in the associated WI

@W-10290520@ https://github.com/forcedotcom/cli/issues/1318
@W-10290524@ https://github.com/forcedotcom/cli/issues/72
@W-5380834@

### DIFF
--- a/src/dataCommand.ts
+++ b/src/dataCommand.ts
@@ -8,7 +8,7 @@
 import { SfdxCommand } from '@salesforce/command';
 import { AnyJson, Dictionary, get, Nullable } from '@salesforce/ts-types';
 import { fs, Messages, Org, SfdxError } from '@salesforce/core';
-import { BaseConnection, ErrorResult, Record as jsForceRecord, SObject } from 'jsforce';
+import { BaseConnection, ErrorResult, Record as jsforceRecord, SObject } from 'jsforce';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore because jsforce doesn't export http-api
 import * as HttpApi from 'jsforce/lib/http-api';
@@ -123,7 +123,7 @@ export abstract class DataCommand extends SfdxCommand {
     return connection;
   }
 
-  public async query(sobject: SObject<unknown>, where: string): Promise<jsForceRecord<unknown>> {
+  public async query(sobject: SObject<unknown>, where: string): Promise<jsforceRecord<unknown>> {
     const queryObject = this.stringToDictionary(where);
     const records = await sobject.find(queryObject, 'id');
     if (!records || records.length === 0) {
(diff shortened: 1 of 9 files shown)


src/dataCommand.ts
```diff
@@ -182,7 +182,11 @@ export abstract class DataCommand extends SfdxCommand {
         throw new Error(messages.getMessage('TextUtilMalformedKeyValuePair', [pair]));
       } else {
         const key = pair.substr(0, eqPosition);
-        constructedObject[key] = this.convertToBooleanIfApplicable(pair.substr(eqPosition + 1));
+        if (pair.includes('{') && pair.includes('}')) {
```
mshanemc, line 185: I could see this being true but not being parseable JSON (ex: a big textArea that's referring to some code like a Case).

So it's probably a fine solution but we'll need a try/catch on the JSON.parse because it might not be, and then we'd want to return the original value?

test/commands/force/data/record/dataRecord.nut.ts
```diff
@@ -247,4 +247,20 @@ describe('data:record commands', () => {
       expect(result).to.have.property('Bool__c', false);
     });
   });
+
+  it('will parse JSON correctly for update', () => {
```
mshanemc, line 251: let's have a test to match the above comment--something that isn't valid json but contains curly brackets.

mshanemc (changes requested):
I like having these new exemplars for the edge cases that keep popping up.

mshanemc (approved):

src/dataCommand.ts
```diff
@@ -137,10 +137,10 @@ export abstract class DataCommand extends SfdxCommand {
       );
     }
 
-    return this.normalize<Record<unknown>>(records);
+    return this.normalize<jsforceRecord<unknown>>(records);
   }
 
-  protected stringToDictionary(str: string): Dictionary<string | boolean> {
+  protected stringToDictionary(str: string): Dictionary<string | boolean | Record<string, unknown>> {
```
shetzel, line 143: For the sake of readability you might consider making a new type for this.

shetzel (approved):

mshanemc:
QA notes

`force:data:soql:query -t -q "select id, Metadata from RemoteProxy"`
✅ has valid output
same, but with the description field being invalid json `"description": "Can I nest some {invalid JSON} here"`
✅ has valid output (treats the invalid json as string)
same, but the description field is actual json `"{"foo":"bar"}"`
✅ escapes it as JSON `"description": "\"{\"foo\":\"bar\"}\""`
same, but with nested JSON and no outer quotes `{"foo":"bar", "nested": {"a": 15}}`
✅ fully escaped `"description": "{\"foo\":\"bar\", \"nested\": {\"a\": 15}}"`

mshanemc:
`force:data:soql:query -q "SELECT Amount, Id, Name, (SELECT Quantity, ListPrice,PriceBookEntry.UnitPrice, PricebookEntry.Name,PricebookEntry.product2.Family FROM OpportunityLineItems) FROM Opportunity"` with setup per WI
✅ 👎🏻 that's an awful looking table, BUT it does have all the values

Screen Shot 2022-02-07 at 12 29 05 PM

PR #285, opened 2022-03-30
fix: bulk batching wide fIles (cli/#1460)

aheber:
### What does this PR do?
Enables splitting files at the 10M character line and not just the 10K row line.

Also added a fix and test so we don't create empty batches if we have exactly 10K rows of input.

https://github.com/forcedotcom/cli/issues/1460

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "@salesforce/ts-types": "^1.5.20",
     "chalk": "^4.1.0",
     "csv-parse": "^4.16.3",
+    "csv-stringify": "^6.0.5",
     "tslib": "^2"
   },
   "devDependencies": {
--- a/src/batcher.ts
+++ b/src/batcher.ts
@@ -10,8 +10,12 @@ import parse = require('csv-parse');
 import { Batch, BatchInfo, BatchResultInfo, JobInfo } from 'jsforce';
 import { UX } from '@salesforce/command';
 import { Job } from 'jsforce/job';
+import { stringify } from 'csv-stringify/sync';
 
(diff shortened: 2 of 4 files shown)


aheber:
Will want to discuss decisions made here. I'm adding a new direct dependency (already an indirect dep of csv-parser so no actual additional packages installed).

This change has performance impact, I think it increased the existing splitting text by 2-3x (~200ms to ~600ms with lots of variation) in my tests on my machine. I don't suspect this is slow enough to tank the change but wanted to call it out.

While I've done research to understand how jsForce was generating the CSVs (same library as I've added) this doesn't guarantee that it will also produce exactly the same CSV as what I'm measuring so the uploaded size could be different, maybe think through any gaps and try and cover them?

package.json
```diff
@@ -89,6 +89,7 @@
     "@salesforce/ts-types": "^1.5.20",
     "chalk": "^4.1.0",
     "csv-parse": "^4.16.3",
+    "csv-stringify": "^6.0.5",
```
shetzel, line 92: jsyk, we need to get approval for all 3rd party libs.  This one looks like it shouldn't be a problem.

shetzel (commented):
Looks good - thanks!  Once we get approval to use `csv-stringify` and manual testing goes well we can merge it.

shetzel:
I tested with a small and large csv and everything bulk uploaded just fine.  I didn't see any significant degradation in performance.  Once the library is approved (should be sometime this week) we can merge this.

aheber:
Great news, appreciate the update.

PR #306, opened 2022-05-26
fix: respect maxQueryLimit config

cristiand391:
### What does this PR do?

Moves config agg logic from core into the plugin, also makes `SoqlQuery` return records emitted by jsforce instead of the full `response`.

Testing:

1. link [command](https://github.com/salesforcecli/command/pull/83) into `plugin-data`

it should respect:
`SFDX_MAX_QUERY_LIMIT` env var
`sfdx config:set maxQueryLimit=1 -g` config

[W-11204187](https://gus.lightning.force.com/a07EE00000xoHkAYAU)

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
   },
   "dependencies": {
     "@oclif/core": "^1.6.4",
-    "@salesforce/command": "^5.1.2",
+    "@salesforce/command": "^5.1.3",
     "@salesforce/core": "^3.19.0",
     "@salesforce/ts-types": "^1.5.20",
     "@types/fs-extra": "^9.0.13",
--- a/yarn.lock
+++ b/yarn.lock
@@ -727,7 +727,7 @@
     is-wsl "^2.1.1"
     tslib "^2.3.1"
 
-"@oclif/core@^1.0.8", "@oclif/core@^1.2.1", "@oclif/core@^1.3.1", "@oclif/core@^1.3.4", "@oclif/core@^1.3.6", "@oclif/core@^1.6.4", "@oclif/core@^1.7.0":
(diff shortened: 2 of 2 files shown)


src/commands/force/data/soql/query.ts
```diff
@@ -33,19 +36,44 @@ const commonMessages = Messages.loadMessages('@salesforce/plugin-data', 'message
...
+    const queryOpts: Partial<QueryOptions> = {
+      autoFetch: true,
+      maxFetch: (config.getInfo('maxQueryLimit').value as number) || 10000,
+    };
+
+    const records: Record[] = [];
+
+    const result = await connection.query(query, queryOpts).on('record', (rec) => records.push(rec));
+
+    const totalSize = getNumber(result, 'totalSize', 0);
```
RodEsp, line 54: It seems to me like there should be some sort of check for the edge case where `result.totalSize` doesn't exist because it we just set it to 0 the message users get will not make any sense.


It'll be something like this:
```
The query result is missing -37892 records due to a 10,000 record limit. Increase the number of records returned by setting the config value "maxQueryLimit" or the environment variable "SFDX_MAX_QUERY_LIMIT" to 0 or greater than 10,000.
```

RodEsp (approved):

cristiand391:
skipped UTs:
✅  `it.skip('should handle a simple query with all records returned in single call'`
❌  `it.skip('should handle a query with a subquery'` this is actually broken in latest-rc, the UT was just checking the json object which is correct, there's a bug in the table code.
Screen Shot 2022-05-26 at 19 38 19
✅  `it.skip('should handle empty query'`
no error,  blank output..

------------------------------------------------------------------------

✅  respects `maxQueryLimit` setting:
✅  json output matches
Screen Shot 2022-05-26 at 19 38 19

✅  respects `SFDX_MAX_QUERY_LIMIT` env var
✅  json output matches
Screen Shot 2022-05-26 at 19 45 22

PR #307, opened 2022-05-31
fix: human output no longer displaying undefined

peternhale:
### What does this PR do?
fixed case for subquery data displaying undefined
fixed case where subquery returns more than one row
@W-11216045@

### DIFF
--- a/src/reporters.ts
+++ b/src/reporters.ts
@@ -124,7 +124,15 @@ export class HumanReporter extends QueryReporter {
         (field: string) =>
           (formattedColumns[field] = {
             header: field.toUpperCase(),
-            get: (row): string => get(row, field) as string,
+            get: (row): string => {
+              // first test if key exists, if so, return value
+              if (Reflect.has(row, field)) {
+                return (Reflect.get(row, field) as string) || '';
+              } else {
+                // if not, try to find it query
+                return get(row, field) as string;
+              }
+            },
           })
       );
     return formattedColumns;
@@ -136,7 +144,7 @@ export class HumanReporter extends QueryReporter {
     // some fields will return a JSON object that isn't accessible via the query (SELECT Metadata FROM RemoteProxy)
     // some will return a JSON that IS accessible via the query (SELECT owner.Profile.Name FROM Lead)
     // querying (SELECT Metadata.isActive FROM RemoteProxy) throws a SOQL validation error, so we have to display the entire Metadata object
-    queryResults.map((qr) => {
+    queryResults.forEach((qr) => {
       const result = qr as Record<string, unknown>;
       this.data.columns.forEach((col) => {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
@@ -167,6 +175,7 @@ export class HumanReporter extends QueryReporter {
           }
         }
 
+        const subResults: Array<Record<string, unknown>> = [];
         if (children.length > 0) {
           const childrenRows: Record<string, unknown> = {};
           children.forEach((child) => {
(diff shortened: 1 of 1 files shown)


src/reporters.ts
```diff
@@ -124,7 +124,15 @@ export class HumanReporter extends QueryReporter {
         (field: string) =>
           (formattedColumns[field] = {
             header: field.toUpperCase(),
-            get: (row): string => get(row, field) as string,
+            get: (row): string => {
```
peternhale, line 127: Modified the get to first check if the field, as given, can be referenced directly, This solves case where property has a name like `foo.bar`. If cannot be directly referenced then delegate to `get`

src/reporters.ts
```diff
@@ -180,19 +189,28 @@ export class HumanReporter extends QueryReporter {
             if (childO) {
               const childRecords = getArray(childO, 'records', []);
               // eslint-disable-next-line @typescript-eslint/no-explicit-any
-              childRecords.forEach((record: unknown) => {
+              childRecords.forEach((record: unknown, index) => {
```
peternhale, line 192: There can be more than one child record from a subquery, Account record can have more than one contact associated to it. For subquery records 2-n a synthetic record needs to be added to the results to produce output like 
```
force:data:soql:query -u myalias -q Select name, (Select FirstName, LastName from Contacts) from Account
 NAME                            CONTACTS.FIRSTNAME CONTACTS.LASTNAME 
 ─────────────────────────────── ────────────────── ───────────────── 
 Test                                                                 
 Sample Account for Entitlements Peter              Hale              
                                 null               Test              
Total number of records retrieved: 2.
```

src/reporters.ts
```diff
@@ -180,19 +189,28 @@ export class HumanReporter extends QueryReporter {
...
                 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                const newResult: Record<string, unknown> = {};
                 Object.entries(record as never).forEach(([key, value]) => {
-                  // merge subqueries with the "parent" so they are on the same row
-                  Reflect.defineProperty(result as Record<string, unknown>, `${child.toString()}.${key}`, {
-                    value: value ? value : chalk.bold('null'),
-                  });
+                  if (!index) {
+                    Reflect.defineProperty(result as Record<string, unknown>, `${child.toString()}.${key}`, {
```
peternhale, line 197: Add subquery data to first query record

src/reporters.ts
```diff
@@ -180,19 +189,28 @@ export class HumanReporter extends QueryReporter {
...
                 });
+                if (index) {
+                  subResults.push(newResult);
+                }
               });
             }
           });
         }
-        newResults.push(result);
+        newResults.push(result, ...subResults);
```
peternhale, line 213: Push parent and additional subquery records into result.

src/reporters.ts
WillieRuemmele: can we add a NUT to cover this case?

WillieRuemmele (approved):

WillieRuemmele:
QA:

in the org, create an Account
add 2 Contacts to that account with the `New Contact` button

```
 ➜  sfdx force:data:soql:query -q "SELECT Name, (SELECT LastName FROM Contacts) FROM Account"
 NAME                            CONTACTS.LASTNAME 
 ─────────────────────────────── ───────────────── 
 Sample Account for Entitlements undefined         
 SF                              undefined         
Total number of records retrieved: 2.
Querying Data... done
➜  dreamhouse-lwc git:(main) ✗  hub:(DevHub) scratch:(test-6t6kwus58zn0@example.com)
 ➜  ../../oss/plugin-data/bin/run force:data:soql:query -q "SELECT Name, (SELECT LastName FROM Contacts) FROM Account"
 NAME                            CONTACTS.LASTNAME 
 ─────────────────────────────── ───────────────── 
 Sample Account for Entitlements                   
 SF                              Ruemmele          
                                 Dom   
```

original query printed `undefined`, the new output will add both Contacts' last name to the query correctly

PR #399, opened 2022-11-11
Sm/sfcommand

mshanemc:
> this is for early feedback/review on the sfCommand changes and the unit testing approach.  It is not "finished"...there's be more changes for sf-styling of commands/flags

### What does this PR do?
plugin-data on sfCommand
no fancy-test
closer to linter/ts rules standard

@W-11912467@


messages/bulk.delete.json
```diff
@@ -1,4 +1,5 @@
 {
+  "summary": "bulk delete records from a csv file",
```
mdonnalley, line 2: Should these files be migrated to markdown?

src/commands/force/data/record/update.ts
```diff
@@ -6,88 +6,94 @@
...
 
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/plugin-data', 'record.update');
 const commonMessages = Messages.loadMessages('@salesforce/plugin-data', 'messages');
 
-export default class Update extends DataCommand {
+export default class Update extends SfCommand<SaveResult> {
+  public static readonly summary = messages.getMessage('summary');
   public static readonly description = messages.getMessage('description');
   public static readonly examples = messages.getMessage('examples').split(os.EOL);
```
mdonnalley, line 21: you can use `messages.getMessages` to avoid using `.split(os.EOL)`

test/commands/force/data/bulk/status.test.ts
```diff
@@ -4,136 +4,196 @@
...
+    eventNames(): Array<string | symbol> {
+      throw new Error('Function not implemented.');
+    },
+  };
+  beforeEach(async () => {
+    await $$.stubAuths(testOrg);
+    await config.load();
+    stdoutSpy = $$.SANDBOX.stub(process.stdout, 'write');
+
+    $$.SANDBOXES.CONNECTION.stub(Bulk.prototype, 'job').returns({
```
mdonnalley, line 108: I _think_ `stubMethod` from @salseforce/ts-types lets you create stubs like this without need to add in `...unusedInterfaceParts`

test/commands/force/data/bulk/status.test.ts
```diff
@@ -4,136 +4,196 @@
...
-      '--jobid',
-      '75054000006ybyHAAQ',
-      '--json',
-    ])
-    .it('will fail with the correct error message', (ctx) => {
-      const result = JSON.parse(ctx.stdout) as StatusResult;
-      expect(result.message).to.equal('Unable to find batch 751540000070Q4RAAU for job 75054000006ybyHAAQ.');
-      expect(result.status).to.equal(1);
-    });
+  const $$ = new TestContext();
```
mdonnalley, line 18: If you do `TestContext({setup: true})`, you don't need to worry instantiating or restoring the defaults stubs: https://github.com/forcedotcom/sfdx-core/blob/main/src/testSetup.ts#L364

test/commands/force/data/bulk/status.test.ts
```diff
@@ -4,136 +4,196 @@
...
+        '751540000070aNsAAI',
+        '--jobid',
+        '75054000006yv68AAA',
+        '--json',
+      ],
+      config
+    );
+
+    // eslint-disable-next-line no-underscore-dangle
+    const result = await cmd._run();
```
mdonnalley, line 180: You might be able to avoid disabling eslint if you do this:

`await Status.run(args, config)`

test/commands/force/data/soql/query/dataSoqlQuery.nut.ts
```diff
@@ -59,30 +60,37 @@ function runQuery(
...
+          executable: 'sfdx',
+          config: 'config/project-scratch-def.json',
+          setDefault: true,
+        },
       ],
       project: { sourceDir: path.join('test', 'test-files', 'data-project') },
+      devhubAuthStrategy: 'AUTO',
     });
+    // get default devhub username
+    hubOrgUsername = execCmd<[{ key: string; value: string }]>('config:get defaultdevhubusername --json', {
```
mdonnalley, line 78: you can access the devhub this way, `testSession.hubOrg`

https://github.com/salesforcecli/cli-plugins-testkit/blob/main/src/testSession.ts#L109

test/commands/force/data/bulk/status.test.ts
mshanemc: pretty sure I tried that for a bit and gave up.  Wasn't sure from the other examples of it in use how to make it work with testSetup, and then the types still weren't right.

test/commands/force/data/bulk/status.test.ts
mshanemc: ooh, we should doc that over in sfdx-core's readme or on one of the /examples/test/unit

test/commands/force/data/bulk/status.test.ts
mshanemc: that does avoid the eslint, but oclif has the result typed as `any`
Screen Shot 2022-11-14 at 9 36 09 AM

I'm gonna convert the ones that don't use the result.

mshanemc:
close in favor of https://github.com/salesforcecli/plugin-data/pull/408

PR #427, opened 2022-12-06
fix: allow single quotes in key=value pairs

RodEsp:
### What does this PR do?
Allows using `'` in the value  of name-value pairs for flags such as `force:data:record:create`'s `--values`.

https://github.com/forcedotcom/cli/issues/1820
@W-12105706@

### DIFF
--- a/src/dataCommand.ts
+++ b/src/dataCommand.ts
@@ -222,48 +222,26 @@ export abstract class DataCommand extends SfdxCommand {
   }
 
   /**
-   * Splits a sequence of 'key=value key="leftValue rightValue"   key=value'
-   * into a list of key=value pairs, paying attention to quoted whitespace.
+   * Splits a sequence of "key=value key=leftValue rightValue   key=value"
+   * into a list of key=value pairs, paying attention to whitespace in values.
+   *
+   * e.g. "key=value key=leftValue rightValue   key=value0"
+   * becomes ["key=value", "key=leftValue rightValue", "key=value"]
    *
    * This is NOT a full push down-automaton so do NOT expect full error handling/recovery.
    *
    * @param {string} text - The sequence to split
    */
   private parseKeyValueSequence(text: string): string[] {
-    const separator = /\s/;
+    // The regex below uses a positive lookbehind and a positive lookahead to split the string into key=value pairs
+    //   regardless of whitespace or single quotes in the values
+    // Use regex101.com and paste this regex to see how it works
+    const matches = text.match(/(\S+)=(?<=['"]?)([\d\w'"\s]+)(?=\s|$)/g);
 
-    let inSingleQuote = false;
-    let inDoubleQuote = false;
-    let currentToken: string[] = [];
-    const keyValuePairs: string[] = [];
-
-    const trimmedText = text.trim();
-    for (const currentChar of trimmedText) {
-      const isSeparator = separator.exec(currentChar);
-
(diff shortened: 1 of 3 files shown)


src/dataCommand.ts
```diff
@@ -222,48 +222,26 @@ export abstract class DataCommand extends SfdxCommand {
...
-          currentToken = [];
-        }
-      } else {
-        currentToken.push(currentChar);
-      }
-    }
+    // The regex below uses a positive lookbehind and a positive lookahead to split the string into key=value pairs
+    //   regardless of whitespace or single quotes in the values
+    // Use regex101.com and paste this regex to see how it works
+    const matches = text.match(/(\S+)=(?<=['"]?)([\d\w'"\s]+)(?=\s|$)/g);
```
peternhale, line 239: @RodEsp the `\d` in `[\d\w'"\s]` is redundant, numbers are covered by `\w`

peternhale: I love all the red in the diff here :)

peternhale (changes requested):
I know there are not a lot UTs in this modules, but maybe one or two more around this change.

src/dataCommand.ts
WillieRuemmele: I tried pasting into regex101, and it had errors?

you can save your regex101 like https://regex101.com/r/gYqdkb/1 (ctrl+s) might be nice to build out an example there and save the url in the comments

peternhale: @WillieRuemmele you need to pick the regex flavor. I chose ECMAScript and it worked for me.

RodEsp:
Heya @peternhale, this is ready for another look. 

Unfortunately, I could not get the regex to work with both weird Unicode characters and values like `key="leftvalue rightvalue"` at the same time, so it looks like we're stuck with the current lengthy code that checks for the existence of `'`s and `"`s.

src/dataCommand.ts
```diff
@@ -238,13 +238,17 @@ export abstract class DataCommand extends SfdxCommand {
     const keyValuePairs: string[] = [];
 
     const trimmedText = text.trim();
+
+    const singleQuoteCount = (trimmedText.match(/'/g) || []).length;
+    const doubleQuoteCount = (trimmedText.match(/"/g) || []).length;
+
     for (const currentChar of trimmedText) {
-      const isSeparator = separator.exec(currentChar);
+      const isSeparator = separator.exec(currentChar) !== null;
```
peternhale, line 246:
```suggestion
      const isSeparator = separator.test(currentChar);
```

peternhale (changes requested):
@RodEsp I found another valid string that does not make it through the function intact, but it probably would of failed in the current code too.
```
  it('should allow double and single quotes in key=value pairs', () => {
    const dict = dataCommand.testStringToDictionary("key='39° 44' 31.3548\" N'");

    expect(dict).to.deep.equal({
      key: '39° 44\' 31.3548" N',
    });
  });
```

RodEsp:
Yeah, that didn't work without the changes in this PR either. The problem is with values that have quoted strings inside quoted strings, e.g. `key="this is a 'quote'"` or `key='this is a "quote"'`. I think it'll be a greater effort to get that to work, but for the sake of fixing https://github.com/forcedotcom/cli/issues/1820 asap I'd like to merge this now and I created https://github.com/forcedotcom/cli/issues/1837.

peternhale (approved):

PR #527, opened 2023-03-26
create data:export:bulk command

AllanOricil:
### What does this PR do?

- it adds the `data:export:bulk` command. This command is able to export all records of an sobject using the Bulk API V2 without throwing  `FATAL Error: Javascript Heap out of Memory`. 

- it removes bulk features from the `data:soql:query` command. I decided to do that because in my opinion it makes more sense to use the bulk api to export data only. If people want to display a LIMITED number of records, they can use normal apis.

# Demo

obs: pay attention to the amount of memory and CPU that is used on each execution. This information is displayed in the VS Code status bar

image

### exporting as json

https://user-images.githubusercontent.com/55927613/227749221-7e9dbebf-3631-40ad-b106-615bdc3b839c.mov

### exporting as csv

https://user-images.githubusercontent.com/55927613/227749198-6052bd4a-6e1b-485b-82e5-37a51209c4fa.mov

### sync json output
image

### async json output
image

https://github.com/forcedotcom/cli/issues/1995

### DIFF
--- a/src/commands/data/query.ts
+++ b/src/commands/data/query.ts
@@ -6,7 +6,7 @@
  */
 
 import * as fs from 'fs';
-import { Connection, Logger, Messages, SfError } from '@salesforce/core';
+import { Connection, Logger, Messages } from '@salesforce/core';
 import { Record } from 'jsforce';
 import {
   AnyJson,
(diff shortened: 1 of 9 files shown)


AllanOricil:
If these changes are accepted, the following items must be done:

TODOS:
- update jsforce dependency package when these changes are released:  https://github.com/jsforce/jsforce/pull/1317
- refactor the data:query:resume. This one can then become `data:export:bulk:resume`, which should either create a new file, or append new results to the file that was created when the initial job was interrupted.
- create a new SF env variable called  SF_BULK_EXPORT_MAX_RECORDS so that Users can overwrite the default value of 10000 records per page.
- create tests

src/commands/data/query.ts
```diff
@@ -57,23 +55,10 @@ export class DataSoqlQueryCommand extends SfCommand<unknown> {
       aliases: ['usetoolingapi'],
       deprecateAliases: true,
     }),
-    bulk: Flags.boolean({
```
WillieRuemmele, line 60: Removing any of these flags will be a breaking change without going through the standard deprecation process... even then, this decision will be made by our PM.

These flags will be required before merging

WillieRuemmele (changes requested):

src/commands/data/query.ts
AllanOricil: ok, when your PM decides to take over this hidden bug, you can contact me again and I will do the changes

PR #594, opened 2023-06-12
Wr/bulk2.0 chunks

WillieRuemmele:
### What does this PR do?
allows users to upload more than the 150mb limit of Bulk 2.0

it splits the file into two smaller batches, and process those individually, and then merges the results

NOTE: the jobId/jobInfo returned in the json is the last batch to be sent

@W-13314448@
https://github.com/forcedotcom/cli/issues/2084

### DIFF
--- a/src/commands/data/upsert/bulk.ts
+++ b/src/commands/data/upsert/bulk.ts
@@ -4,8 +4,11 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { Messages } from '@salesforce/core';
+import * as fs from 'fs';
+import * as os from 'os';
+import { Connection, Messages } from '@salesforce/core';
 import { Flags } from '@salesforce/sf-plugins-core';
+import { Duration } from '@salesforce/kit';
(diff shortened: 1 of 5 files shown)


src/commands/data/upsert/bulk.ts
```diff
@@ -34,6 +36,28 @@ export default class Upsert extends BulkOperationCommand {
...
 
     await validateSobjectType(flags.sobject, conn);
+    // if the file is >150 mb, break it up into chunks, we'll deploy those individually
+    const fileStat = fs.statSync(flags.file);
+    if (fileStat.size >= 150000000) {
+      const batches = writeBatches(fs.readFileSync(flags.file, { encoding: 'utf8' }));
+
+      const results = await Promise.all(
+        batches.map(async (batch, i) =>
+          this.runBulkOperation(flags.sobject, `batch${i}.csv`, conn, flags.async ? 0 : flags.wait?.minutes, 'upsert', {
```
shetzel, line 46: nit: don't need to compute wait time with each batch.  It's the same for each one so we can compute outside of the loop.  We can also reuse it for the < 150MB code path.

shetzel (approved):
Approved with a nit comment

PR #615, opened 2023-07-04
feat: add verbose flag to bulk delete and upsert

R0Wi:
### What does this PR do?

This adds a `--verbose` flag to the `sf data delete bulk` and `sf data upsert bulk` commands. If set, failed records will be printed as table:

image

https://github.com/forcedotcom/cli/issues/2221

### Discussion

* How to write tests for this? I tried to mock the `node-fetch` library which is used by `BulkApiV2.request` but this is quite hard to accomplish. Maybe someone has some other ideas?
* Currently we skip the table printing if `--json` is set. This is because otherwise the printed JSON would be malformed.
* How to deal with the returned `Id` column? The only thing we know for sure is that SF will give us the `sf__Id` column. But if an error occures (for example when trying to delete records) this column won't be filled and it might make more sense to have the actual source `Id` column (or external id in case of upsert) printed here. This might end in a different implementation for `upsert` and `delete`.

### DIFF
--- a/src/bulkOperationCommand.ts
+++ b/src/bulkOperationCommand.ts
@@ -58,12 +58,10 @@ export abstract class BulkOperationCommand extends BulkBaseCommand {
     async: Flags.boolean({
       char: 'a',
       summary: messages.getMessage('flags.async.summary'),
-      default: false,
       exclusive: ['wait'],
     }),
     verbose: Flags.boolean({
       summary: messages.getMessage('flags.verbose'),
-      default: false,
     }),
   };
 
@@ -98,15 +96,16 @@ export abstract class BulkOperationCommand extends BulkBaseCommand {
     return job.check();
   }
 
-  private printBulkErrors(failedResults: IngestJobV2FailedResults<Schema>): void {
+  private static printBulkErrors(failedResults: IngestJobV2FailedResults<Schema>): void {
     const columns = {
       id: { header: 'Id' },
+      sfId: { header: 'Sf_Id' },
       error: { header: 'Error' },
     };
     const options = { title: `Bulk Failures [${failedResults.length}]` };
     ux.log();
     ux.table(
-      failedResults.map((f) => ({ id: f.sf__Id, error: f.sf__Error })),
+      failedResults.map((f) => ({ id: 'Id' in f ? f.Id : '', sfId: f.sf__Id, error: f.sf__Error })),
       columns,
       options
     );
@@ -156,17 +155,17 @@ export abstract class BulkOperationCommand extends BulkBaseCommand {
         if (!isBulkV2RequestDone(jobInfo)) {
           return result;
         }
-        let records = null;
-        if (verbose) {
-          records = await this.job.getAllResults();
+        if (this.jsonEnabled()) {
+          result.records = transformResults(await this.job.getAllResults());
+        }
+        // We only print human readable error outputs if --json is not specified.
+        // The JSON result itself will already contain the error information (see above).
+        else if (verbose) {
+          const records = await this.job.getAllResults();
           if (records?.failedResults?.length > 0) {
-            this.printBulkErrors(records.failedResults);
+            BulkOperationCommand.printBulkErrors(records.failedResults);
           }
         }
-        if (!this.jsonEnabled()) {
-          return result;
-        }
-        result.records = transformResults(records ?? await this.job.getAllResults());
         return result;
       } catch (err) {
         this.spinner.stop();
--- a/src/bulkOperationCommand.ts
+++ b/src/bulkOperationCommand.ts
@@ -10,7 +10,8 @@ import * as os from 'os';
 import { Flags } from '@salesforce/sf-plugins-core';
 import { Duration } from '@salesforce/kit';
 import { Connection, Messages } from '@salesforce/core';
-import { BulkOperation, IngestJobV2, IngestOperation, JobInfoV2 } from 'jsforce/api/bulk';
+import { ux } from '@oclif/core';
+import { BulkOperation, IngestJobV2, IngestJobV2FailedResults, IngestOperation, JobInfoV2 } from 'jsforce/api/bulk';
 import { Schema } from 'jsforce';
 import { orgFlags } from './flags';
 import { BulkDataRequestCache } from './bulkDataRequestCache';
@@ -60,6 +61,10 @@ export abstract class BulkOperationCommand extends BulkBaseCommand {
       default: false,
       exclusive: ['wait'],
     }),
+    verbose: Flags.boolean({
+      summary: messages.getMessage('flags.verbose'),
+      default: false,
+    }),
   };
 
   /**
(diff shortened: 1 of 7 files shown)


src/bulkOperationCommand.ts
```diff
@@ -60,6 +61,10 @@ export abstract class BulkOperationCommand extends BulkBaseCommand {
       default: false,
       exclusive: ['wait'],
     }),
+    verbose: Flags.boolean({
+      summary: messages.getMessage('flags.verbose'),
+      default: false,
```
WillieRuemmele, line 66: the default value for Boolean flags is `false` can you remove this line?

WillieRuemmele:
Hi @R0Wi - thanks for the contribution - the first glance looks pretty good. a few minor things that can easily be addressed. I'll pull down your fork tomorrow and click through things more in depth then and leave my comments.

For the testing question, I would suggest writing a few NUTs (not unit tests) for this, checkout the `test/commands/data/dataBulk.nut.ts` tests, you'll maybe want to add something to the `test-files` directory as well to create a reproducible test, it would be great to have some sample data that fails (if that's possible) so we can verify the table data is correct, and verify that the `--json` flag returns correctly formed output.

For the id question... would it make sense to include both of the IDs? maybe one as `sf__Id` and one as `Id` for when we error? It would be better to keep them separated to create a consistent behavior and so you don't get different `Id`s in the same field based on what happened during the command execution

src/bulkOperationCommand.ts
```diff
@@ -93,11 +98,26 @@ export abstract class BulkOperationCommand extends BulkBaseCommand {
     return job.check();
   }
 
+  private printBulkErrors(failedResults: IngestJobV2FailedResults<Schema>): void {
+    const columns = {
+      id: { header: 'Id' },
+      error: { header: 'Error' },
+    };
+    const options = { title: `Bulk Failures [${failedResults.length}]` };
+    ux.log();
```
WillieRuemmele, line 107: you can create a "json aware ux" with `const ux = new Ux({jsonEnabled: this.jsonEnabled()});` You might need to move this so it has access to the command's `json` flag

src/bulkOperationCommand.ts
```diff
@@ -133,10 +153,20 @@ export abstract class BulkOperationCommand extends BulkBaseCommand {
         }
         this.displayBulkV2Result(jobInfo);
         const result = { jobInfo } as BulkResultV2;
-        if (!isBulkV2RequestDone(jobInfo) || !this.jsonEnabled()) {
+        if (!isBulkV2RequestDone(jobInfo)) {
+          return result;
+        }
+        let records = null;
+        if (verbose) {
```
WillieRuemmele, line 160: could this be moved to `displayBulkV2Result`?

R0Wi: It could be moved, yes. But then we'd need to change the signature of the `BulkBaseCommand` function to something like this: `async displayBulkV2Result(jobInfo: JobInfoV2, verbose: boolean = false): Promise<void>` because
* We might need to fetch results in that function (`async`)
* Function is also used in `ResumeBulkCommand.resume`, which does not have a `verbose` flag (so adding a default `boolean` parameter for `verbose`)

Also we'd then need to move `printBulkErrors` to the `BulkBaseCommand` class. Not sure if that makes sense since we're defining the parameter in the descendant `BulkOperationCommand`?

And finally, we might pull the results twice (`await this.job.getAllResults()`) if the `verbose` flag is true. One time inside of the `BulkBaseCommand.displayBulkV2Result` and one time again in this function here (`BulkOperationCommand.runBulkOperation`).

So maybe something like that would be better:

```js
        this.displayBulkV2Result(jobInfo);
        const result = { jobInfo } as BulkResultV2;
        if (!isBulkV2RequestDone(jobInfo)) {
          return result;
        }
        if (!this.jsonEnabled()) {
          if (verbose) {
            const records = await this.job.getAllResults();
            if (records?.failedResults?.length > 0) {
              this.printBulkErrors(records.failedResults);
            }
          }
          return result;
        }
        result.records = transformResults(await this.job.getAllResults());
        return result;
```

src/bulkOperationCommand.ts
R0Wi: Sure, will remove this in the next commit 👍

src/bulkOperationCommand.ts
R0Wi: I can change this, of course. But since the Ux should never be printed if `this.jsonEnabled()` is `true` (because the error results are already part of the returned JSON itself), I think we can assume that we'd really like to print a human readable table here?

R0Wi:
>  so if you want, if you provide the data, commands, and expected behavior, I could write the tests too

That sounds great! For my personal tests I used the following setup:

**[1] Testcase "should print table because `--verbose` flag is set and there are errors"**

* Create a new file `delete.csv`:

```csv
Id
<Some non-existing Id>
``` 

* Run `sf data bulk delete --file delete.csv --sobject Account --verbose`
* The SF API `services/data/v58.0/jobs/ingest/<JobId>/failedResults/` will return a CSV like this:

```csv
"sf__Id","sf__Error",Id
"","MALFORMED_ID:malformed id <ID>:--","<ID>"
```

**[2] Testcase "should not print table because `--verbose` flag is set but there are no errors"**

* Insert a new `Account` in Salesforce and keep track of the Id
* Create a new file `delete.csv`:

```csv
Id
<Tracked Id of Account to be deleted>
``` 

* Run `sf data bulk delete --file delete.csv --sobject Account --verbose`

**Other Testcases can be setup by combining the above approach with `--verbose` and `--json` flags set/unset** 

Please let me know if you need further information.

----

>  would it make sense to include both of the IDs?

Yes, makes sense. Depending on the concrete operation the `sf__Id` might not be filled (for example if a bulk delete for a record failed). On the other hand we might need to check if the result contains an `Id` column or not, think that depends on the data you uploaded to SF (see above example). So we can add an `Id` column and fill values if there are any?

WillieRuemmele:
Awesome @R0Wi - I've added those tests, that in theory, following your test scenarios, should pass... they might need to be tweaked based on some changes you make... I did notice that the table is being printed in `--json` scenarios, try 
```bash
 ➜  ../../oss/rowi-data/bin/dev data:delete:bulk --file data.csv --sobject Account --verbose --wait 10 --json

Bulk Failures [1]
====================================================
| Id Error                                           
| ── ─────────────────────────────────────────────── 
|    MALFORMED_ID:malformed id 123e1232133nugGQAQ:-- 
{
  "status": 1,
  "result": {
    "jobInfo": {
```

so it seems like `ux` isn't json aware unless explicitly told by the flag

R0Wi:
Great, thank you! Yes I already noticed that. Will commit a fix tomorrow :-)

R0Wi:
Hi @WillieRuemmele, I adjusted the code a bit and "manually" tested the testcases. They are now working for me. Think that's it from my end at the moment, any feedback appreciated 🚀

WillieRuemmele:
all tests passed and validated, thanks for the contribution, this will be in the `nightly` build tonight, `latest-rc` next wednesday, and `latest` the wednesday after that. 

Thanks for your contribution @R0Wi 🏆 ❤️

PR #697, opened 2023-09-25
fix: assert flag type

mdonnalley:
### What does this PR do?

Assert `target-org` flag type so that `@oclif/core@v3` won't think that it's possibly undefined. 

Using the spread operator to extend flags is an anti-pattern and should be avoided. The types in v3 are more accurate which means we can no longer get away with anti-patterns like this. If we have to do this sort of thing the better approach would be to expose the defaults for the custom flag so that they can be reused.

So instead of this,

```typescript
// @salesforce/sf-plugins-core
export const requiredHubFlagWithDeprecations = requiredHubFlag({
  aliases: ['targetdevhubusername'],
  deprecateAliases: true,
  required: true,
});
```

```typescript
// @salesforce/plugin-data
'target-org': {...requiredHubFlagWithDeprecations, summary: 'a summary'}
```

You would do this,

```typescript
// @salesforce/sf-plugins-core
export const requiredHubFlagWithDeprecationsDefaults = {
  aliases: ['targetdevhubusername'],
  deprecateAliases: true,
  required: true,
};
```

```typescript
// @salesforce/plugin-data
'target-org': Flags.custom<Org>(requiredHubFlagWithDeprecationsDefaults)(
  { summary: 'a summary' },
)
```

[skip-validate-pr]

### DIFF
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -4,7 +4,8 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { Messages } from '@salesforce/core';
+import { Interfaces } from '@oclif/core';
+import { Messages, Org } from '@salesforce/core';
 import {
   Flags,
   loglevel,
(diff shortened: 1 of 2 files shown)


src/flags.ts
```diff
@@ -24,8 +25,17 @@ export const perflogFlag = Flags.boolean({
...
+// custom flag is the preferred approach. But if it must be done,
+// then the type must be asserted so that @oclif/core can properly
+// parse the flag's type.
+const targetOrg = {
+  ...requiredOrgFlagWithDeprecations,
+  summary: messages.getMessage('flags.targetOrg.summary'),
+} as Interfaces.OptionFlag<Org>;
+
 export const orgFlags = {
-  'target-org': { ...requiredOrgFlagWithDeprecations, summary: messages.getMessage('flags.targetOrg.summary') },
```
mshanemc, line 28: I don't think the summary in this plugin `Org alias or username to use for the target org.` needs to exist.

https://github.com/salesforcecli/plugin-data/blob/main/messages/messages.md#flagstargetorgsummary

since the default is `Username or alias of the target org.`
https://github.com/salesforcecli/sf-plugins-core/blob/main/messages/messages.md#flagstargetorgsummary

mshanemc (approved):

PR #705, opened 2023-10-04
fix(bulk v2): catch job errors while polling

cristiand391:
### What does this PR do?
requires https://github.com/jsforce/jsforce/pull/1367 

when polling for an ingest job status like `await job.poll()` jsforce will emit a failure event if the job fails but it doesn't throw so all job errors during polling are swallowed.
This PR adds a handler for the `failed` event and make `data` commands throw.

`failed` event: https://github.com/jsforce/jsforce/blob/8a359d663215081becc63561735b947df759c185/src/api/bulk.ts#L1505

@W-13502104@


shetzel (approved):

src/batcher.ts
```diff
@@ -141,7 +141,7 @@ export class Batcher {
                 err.message = messages.getMessage('ExternalIdRequired', [sobjectType]);
                 job.emit('error', err);
               }
-              if (err.message.startsWith('Polling time out')) {
+              if (err.name === 'PollingTimeout') {
```
cristiand391, line 144: this is from the bulk v1 API:
https://github.com/jsforce/jsforce/blob/7c8e6a130a149cc7f310e6628f786f94a98c91b9/src/api/bulk.ts#L707
https://github.com/jsforce/jsforce/blob/7c8e6a130a149cc7f310e6628f786f94a98c91b9/src/api/bulk.ts#L484

src/commands/data/query.ts
```diff
@@ -150,7 +150,7 @@ export class DataSoqlQueryCommand extends SfCommand<unknown> {
       return transformBulkResults(res, query);
     } catch (e) {
       const err = e as Error & { jobId: string };
-      if (timeout.minutes === 0 && err.message.includes('Polling time out')) {
+      if (timeout.minutes === 0 && err.name === 'JobPollingTimeout') {
```
cristiand391, line 153: bulk v2 query job, thrown from here:
https://github.com/jsforce/jsforce/blob/7c8e6a130a149cc7f310e6628f786f94a98c91b9/src/api/bulk.ts#L1249
name: https://github.com/jsforce/jsforce/blob/7c8e6a130a149cc7f310e6628f786f94a98c91b9/src/api/bulk.ts#L498

PR #763, opened 2023-12-07
fix: default wait

mshanemc:
> there's a bug in plugins-core that broke the NUTs. fix: https://github.com/salesforcecli/sf-plugins-core/pull/470

### What does this PR do?
use defaultValue on Duration flags

https://salesforce-internal.slack.com/archives/G02K6C90RBJ/p1701888844122029
@W-14625639@

QA:
if you want to link the plugin or use bin/run, you need to run `yarn oclif manfiest` first to get the manifest, which is where you could see the problem before this fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "@oclif/core": "^3.14.0",
     "@salesforce/core": "^6.4.0",
     "@salesforce/kit": "^3.0.15",
-    "@salesforce/sf-plugins-core": "^5.0.7",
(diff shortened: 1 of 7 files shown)


mdonnalley (approved):

mdonnalley:
QA

🟢 `wait` flag shows expected value in `help` when manifest is present
🟢 `wait` flag shows expected value in `help` when manifest is not present

mdonnalley (approved):

PR #772, opened 2023-12-14
fix(deps): missing dependency

mshanemc:
https://github.com/forcedotcom/eslint-config-salesforce-typescript/actions/runs/7214346950/job/19656240784

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "@salesforce/sf-plugins-core": "^5.0.10",
     "@salesforce/ts-types": "^2.0.9",
     "chalk": "^5.3.0",
+    "change-case": "^5.3.0",
     "csv-parse": "^4.16.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -114,6 +114,7 @@
     "@salesforce/sf-plugins-core": "^5.0.10",
     "@salesforce/ts-types": "^2.0.9",
     "chalk": "^5.3.0",
+    "change-case": "^5.3.0",
```
cristiand391, line 117: I see `change-case` v4 in the current lockfile, can't find the changelog in the repo
https://github.com/blakeembrey/change-case

Screenshot 2023-12-15 at 12 22 06

cristiand391: we didn't find any breaking change that affects us going v4 -> v5.

PR #844, opened 2024-02-29
Sm/v3-refactoring-party

mshanemc:
### What does this PR do?
- continued refactoring work for jsforce3 (esp async query to not return results or check even once)
- use jsforce3 bulk2 polling events
- allow --use-tooling-api with --bulk on query (https://github.com/forcedotcom/cli/discussions/2667)
- reorganize the reporters, mostly using functions.  If I created a bug, it's somewhere in there
- more UT for reporter logic
- 1 nut fix where bulk delete resume human output wasn't being tested 
- drop support for "sfdx" and ":" in help/message output.  only sf and spaces 

##
there are some // ts-expect-error because `jsforce-node.Connection` isn't **exactly** identical to `sfdx-core.Connection` that extends jsforce v2 

one obvious difference is that they have a different implementation of `conn.bulk/.bulk2` but there are others about private members (ex: sfdx-core inserts its own logger)

When jsforce-node is rolled out in sfdx-core and sfCommand those will start failing, hopefully, and we can remove them

### DIFF
--- a/messages/bulk.base.command.md
+++ b/messages/bulk.base.command.md
@@ -4,12 +4,12 @@ Bulk %s request %s started successfully
 
 # checkStatus
 
-Run command %s data %s resume -i %s -o %s to check status.
+Run command sf data %s resume -i %s -o %s to check status.
 
 # checkJobViaUi
 
 To review the details of this job, run:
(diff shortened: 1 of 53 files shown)


mshanemc:
closing, merging into the main v3 PR

PR #922, opened 2024-05-15
feat: file upload

mshanemc:
### What does this PR do?
adds a new command to upload files and optionally attach them to records
to replace a 3PP plugin command used by one of the commerce plugins

used `create` because of definition here https://github.com/salesforcecli/cli/wiki/Design-Guidelines-Common-Actions

part of https://github.com/forcedotcom/cli/discussions/2344
related to https://github.com/forcedotcom/cli/discussions/2346

### DIFF
--- a/command-snapshot.json
+++ b/command-snapshot.json
@@ -1,5 +1,13 @@
 [
   {
+    "alias": [],
+    "command": "data:create:file",
+    "flagAliases": [],
+    "flagChars": ["f", "i", "n", "o"],
+    "flags": ["api-version", "file", "flags-dir", "json", "name", "parent-id", "target-org"],
+    "plugin": "@salesforce/plugin-data"
+  },
+  {
     "alias": ["force:data:record:create"],
     "command": "data:create:record",
     "flagAliases": ["apiversion", "sobjecttype", "targetusername", "u", "usetoolingapi"],
--- a/command-snapshot.json
+++ b/command-snapshot.json
@@ -3,8 +3,8 @@
     "alias": [],
     "command": "data:create:file",
     "flagAliases": [],
-    "flagChars": ["f", "i", "n", "o"],
-    "flags": ["api-version", "file", "flags-dir", "json", "name", "parent-id", "target-org"],
+    "flagChars": ["f", "i", "o", "t"],
+    "flags": ["api-version", "file", "flags-dir", "json", "parent-id", "target-org", "title"],
     "plugin": "@salesforce/plugin-data"
   },
   {
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
         "longDescription": "Use the data commands to manipulate records in your org. Commands are available to help you work with various APIs. Import CSV files with the Bulk API V2. Export and import data with the SObject Tree Save API. Perform simple CRUD operations on individual records with the REST API.",
         "subtopics": {
           "create": {
-            "description": "Create a record."
+            "description": "Create a record or a file"
           },
           "delete": {
             "description": "Delete a single record or multiple records in bulk."
@@ -70,7 +70,8 @@
           "upsert": {
             "description": "Upsert many records."
           }
-        }
+        },
+        "external": true
       }
     },
     "flexibleTaxonomy": true,
@@ -115,13 +116,15 @@
     "chalk": "^5.3.0",
     "change-case": "^5.4.4",
     "csv-parse": "^4.16.3",
-    "csv-stringify": "^6.4.6"
+    "csv-stringify": "^6.4.6",
+    "form-data": "^4.0.0"
   },
   "devDependencies": {
     "@oclif/plugin-command-snapshot": "^5.1.9",
     "@salesforce/cli-plugins-testkit": "^5.3.4",
     "@salesforce/dev-scripts": "^9.1.1",
     "@salesforce/plugin-command-reference": "^3.0.83",
+    "@salesforce/types": "^1.1.0",
     "eslint-plugin-sf-plugin": "^1.18.3",
     "oclif": "^4.10.2",
     "ts-node": "^10.9.2",
--- a/src/api/file/fileToContentVersion.ts
+++ b/src/api/file/fileToContentVersion.ts
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { readFile } from 'node:fs/promises';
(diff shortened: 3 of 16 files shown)


src/commands/data/create/file.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export default class DataCreateFile extends SfCommand<ContentVersion> {
+  public static readonly summary = messages.getMessage('summary');
+  public static readonly description = messages.getMessage('description');
+  public static readonly examples = messages.getMessages('examples');
+
+  public static readonly flags = {
+    'target-org': Flags.requiredOrg(),
+    'api-version': Flags.orgApiVersion(),
+    name: Flags.string({
```
jshackell-sfdc, line 29: How about calling this flag "--new-name"?  Just to make it clear that, well, the name is new :).  Just a thought.

mshanemc: what's the alternative interpretation that `-new-name` clears up by eliminating?

jshackell-sfdc: I think when i wrote my comment I was thinking that --name on a command is typically used to identify something that already exists; but I just looked at all our commands that have --name, and it's to give something a name, like in this instance.  So ignore, --name is great!

jshackell-sfdc (approved):

package.json
```diff
@@ -47,7 +47,7 @@
         "longDescription": "Use the data commands to manipulate records in your org. Commands are available to help you work with various APIs. Import CSV files with the Bulk API V2. Export and import data with the SObject Tree Save API. Perform simple CRUD operations on individual records with the REST API.",
         "subtopics": {
           "create": {
-            "description": "Create a record."
+            "description": "Create a record or a file"
```
jshackell-sfdc, line 50:
```suggestion
            "description": "Create a record or a file."
```
```suggestion
            "description": "Create a record or a file"
```

command-snapshot.json
```diff
@@ -1,4 +1,12 @@
 [
+  {
+    "alias": [],
+    "command": "data:create:file",
+    "flagAliases": [],
+    "flagChars": ["f", "i", "n", "o"],
+    "flags": ["api-version", "file", "flags-dir", "json", "name", "parent-id", "target-org"],
```
VivekMChawla, line 7: @mshanemc - Is `name` meant to be an API or Developer Name, or more like a label?

mshanemc: it's the name of the file once it gets uploaded to the org.  See the `examples`  where maybe the filename sucks and you want a better name to show up on the Files tab or the File related list.

You can see the effect of that in yesterday's demo https://drive.google.com/file/d/1U4UPL-bnRKU6D1paUmqVBoIIKLvWX4MP/view around 10:00

src/commands/data/create/file.ts
VivekMChawla: Thanks, @mshanemc for referring me back to the demo video regarding the `--name` flag.  That was very helpful.

I see where @jshackell-sfdc was coming from with the `--new-name` suggestion. I think she has a point. For me, it wasn't immediately apparent that setting `--name` was optional and would overwrite the local filename.

Am I correct in understanding that a successful file upload is the combination of the following?

1. Create a `ContentDocument` 
2. Create a `ContentDocumentLink`

If that's correct, is it possible that the `Title` field of the [ContentDocument](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_contentdocument.htm) metadata type is the thing that is getting set when the file name is "overridden" by the `--name`flag?

If that's the case, maybe this flag should be `--title` instead of `--name`?

command-snapshot.json
jshackell-sfdc: In the UI, the name is actually labeled "Title":

image

src/commands/data/create/file.ts
jshackell-sfdc: Ha, funny -- I just realized that now, see my comment/screenshot below!  I like --title

command-snapshot.json
VivekMChawla: @mshanemc - I hadn't noticed this comment from @jshackell-sfdc while I was writing my [other comment](https://github.com/salesforcecli/plugin-data/pull/922#discussion_r1603925110).

I think this is another good reason for changing the `--name` flag to `--title`.

src/commands/data/create/file.ts
mshanemc: > Am I correct in understanding that a successful file upload is the combination of the following?
> 
> 1. Create a `ContentDocument`
> 2. Create a `ContentDocumentLink`

Not exactly.  
- If you're just uploading a file, that's a ContentVersion (which automatically creates its parent ContentDocument)
- the ContentDocumentLink is only used if you pass `--parent-id` that's what related the File to the Record
 
> If that's the case, maybe this flag should be `--title` instead of `--name`?
OK, title it is!

mshanemc: fyi, the title also becomes the filename in the org (for example, if you open it or download it, etc).

VivekMChawla (approved):

package.json
```diff
@@ -115,13 +116,15 @@
     "chalk": "^5.3.0",
     "change-case": "^5.4.4",
     "csv-parse": "^4.16.3",
-    "csv-stringify": "^6.4.6"
+    "csv-stringify": "^6.4.6",
+    "form-data": "^4.0.0"
```
shetzel, line 120: 3PP approved?

src/api/file/fileToContentVersion.ts
```diff
@@ -0,0 +1,49 @@
...
+    url: '/sobjects/ContentVersion',
+    headers: { ...form.getHeaders() },
+    body: form.getBuffer(),
+    method: 'POST',
+  });
+
+  const result = await conn.query<ContentVersion>(
+    `Select Id, ContentDocumentId, Title, FileExtension from ContentVersion where Id='${CV.id}'`
+  );
+  return result.records[0];
```
shetzel, line 48: Is a record guaranteed to be returned from the query?  I don't know the `/sobjects/ContentVersion` API so maybe it can be assumed.  But if it returns a non http 200 the code will continue and either fail during the query or throw a TypeError.

package.json
mshanemc: used by jsforce

src/api/file/fileToContentVersion.ts
mshanemc: it's the ID from one we just created in line 38.

http errors  would throw from conn.query (before line 48).  I can put some checks around it (or switch to singleRecordQuery, ex: CV saves but doesn't show up in the query for some reason)

package.json
mshanemc: also, yes.

shetzel (approved):

PR #967, opened 2024-06-18
Wr/timeout message

WillieRuemmele:
### What does this PR do?
refactors similar util files into one
refactors similar types files into one

catches a timeout error so that the timeout event listener can properly handle, and print instructions

@W-16010988@


src/bulkOperationBase.ts
```diff
@@ -185,7 +187,14 @@ const executeBulkV2DataRequest = async <J extends Schema>(
   job.emit('jobProgress', { remainingTime: remainingTime(Date.now())(endWaitTime), stage: 'uploadComplete' });
   await job.close();
   if (endWaitTime && Date.now() < endWaitTime) {
-    await job.poll(POLL_FREQUENCY_MS, remainingTime(Date.now())(endWaitTime));
+    try {
```
WillieRuemmele, line 190: catching this error, fixes the bug

jshackell-sfdc (changes requested):
see my suggestions

mdonnalley (approved):

mdonnalley:
QA: Willie and decided offline that the NUTs are sufficient QA for this PR.

PR #975, opened 2024-06-27
feat!: ga new tree commands

mshanemc:
### What does this PR do?
GA the new tree commands
See the plan at https://github.com/salesforcecli/plugin-data/blob/main/NewTreeCommandsPlan.md

- update the plan
- do the steps on the plan

unrelated
- the `data resume` command was incorrectly marked deprecated, pointing people to use the old `force:` command.  🤷🏻‍♂️ 

### why aren't there UT for the new commands?

- they're simple (the functions they call have the UT)
- NUTs are better for org import/export with an org, fs, and os issues 

https://github.com/forcedotcom/cli/issues/2738
[@W-16113657@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001vStkzYAC/view)


messages/tree.export.md
```diff
@@ -30,7 +30,7 @@ Directory in which to generate the JSON files; default is current directory.
 
 - Export records retrieved with the specified SOQL query into a single JSON file in the current directory; the command uses your default org:
 
-  <%= config.bin %> <%= command.id %> --query "SELECT Id, Name, (SELECT Name, Address__c FROM Properties__r) FROM Broker__c"
```
WillieRuemmele, line 33: bump dev-scripts to avoid linting the .md files?

src/commands/data/import/tree.ts
```diff
@@ -6,93 +6,54 @@
...
-    'config-help': Flags.boolean({
-      summary: messages.getMessage('flags.config-help.summary'),
-      aliases: ['confighelp'],
-      deprecateAliases: true,
-      hidden: true,
-      deprecated: { message: messages.getMessage('flags.config-help.deprecation') },
-    }),
   };
 
-  public async run(): Promise<ImportResult[] | JsonMap> {
```
WillieRuemmele, line 60: why are we changing so much of these commands for the GA? shouldn't this PR just be changing command files/states/warnings?

WillieRuemmele (changes requested):

src/commands/data/import/tree.ts
mshanemc: the command that was `src/commands/data/import/tree.ts` moved to `src/commands/data/import/legacy/tree.ts` and the command that was `src/commands/data/import/beta/tree.ts` moved here.

git doesn't know we swapped the files so it tries to diff this file

messages/tree.export.md
mshanemc: aww, it's on current dev-scripts.  I don't know why that's still happening.  Maybe something in vscode ext?

jshackell-sfdc:
FYI, I created a separate PR with my messages updates. I thought I was gonna have more comments, which is why I separated it out, but it turned out to be rather small: https://github.com/salesforcecli/plugin-data/pull/976

src/commands/data/export/legacy/tree.ts
```diff
@@ -46,15 +47,19 @@ export default class Export extends SfCommand<DataPlanPart[] | SObjectTreeFileCo
   };
 
   public async run(): Promise<DataPlanPart[] | SObjectTreeFileContents> {
+    this.info(
+      'Try the the new "sf data export beta tree" command.  It support SOQL queries with up to 5 levels of objects!'
```
WillieRuemmele, line 51: if they run the legacy command, you'll get a suggestion for a beta command, it should just b e"data export tree" right
```bash
 ➜  ../../oss/plugin-data/bin/run.js data export legacy tree --query "SELECT ID FROM Account"
Try the the new "sf data export beta tree" command.  It support SOQL queries with up to 5 levels of objects!
```

WillieRuemmele: even then, 
```bash
 ➜  ../../oss/plugin-data/bin/run.js data export beta tree --query "SELECT ID FROM Account"   
 ›   Error: command data:export:beta:tree not found
```

mshanemc: I'm just going to remove that.  at this point, you're intentionally choosing this old command to avoid the new one.

mshanemc: also, you found that I had the "beta" alias wrong on the new commands.

WillieRuemmele (commented):
the beta deprecation is accurate, but there's no information on the legacy commands
```bash
 ➜  ../../oss/plugin-data/bin/run.js data export beta tree --query "SELECT ID FROM Account"
Warning: The "data export beta tree" command has been deprecated. Use "data export tree" instead.
wrote 1 records to Account.json
➜  dreamhouse-lwc git:(main) ✗  hub:(GLOBAL - DevHub) scratch:(test-enhclzordbdi@example.com)
 ➜  ../../oss/plugin-data/bin/run.js data export tree --query "SELECT ID FROM Account" 
wrote 1 records to Account.json
➜  dreamhouse-lwc git:(main) ✗  hub:(GLOBAL - DevHub) scratch:(test-enhclzordbdi@example.com)
 ➜  ../../oss/plugin-data/bin/run.js data export legacy tree --query "SELECT ID FROM Account"       
Wrote 1 records to Account.json
```

WillieRuemmele:
### QA Notes
---

✅ : warning on legacy
✅ : no message on 'data export tree'
✅ : deprecation on 'beta' command
✅ : `--json`

---

💡 : could improve, ok since it matches current behavior -  I didn't have required information in the `Account.json` file
```bash
 ➜  ../../oss/plugin-data/bin/run.js data import tree --files Account.json -o test-ddwy6vblev4p@example.com
Error (ERROR_HTTP_400): {"hasErrors":true,"results":[{"referenceId":"AccountRef1","errors":[{"statusCode":"REQUIRED_FIELD_MISSING","message":"Required fields are missing: [Name]","fields":["Name"]}]}]}

```

✅ : no note to try beta on standard command
✅ : beta is deprecated
✅ : legacy mentions deprecation day
✅ : --json consistent (for my use case)

WillieRuemmele (approved):

PR #1026, opened 2024-08-08
write help for new "data search" command

jshackell-sfdc:
### What does this PR do?

Embellishes the skeleton `--help` for the new "data search" command.

@W-16448626@


jshackell-sfdc:
@WillieRuemmele , note that the `--result-format csv` flag writes the results to a file, its name based on the object that's being queried.  This behavior differs from `data query --result-format csv` which just displays the results to the terminal, but in comma-separated value format. 

I assume that you'll change the behavior of `data search` to match the behavior of `data query`.  If, however, you don't, please let me know so I can change the example and the help for the flag!  Cheers.

PR #1085, opened 2024-10-09
fix: bump oclif/mso

cristiand391:
### What does this PR do?

bump oclif/mso, update calls to `ms.stop` (no longer takes an error).

no function changes, example trying to export records with a bad query:
Screenshot 2024-10-09 at 3 56 42 PM

[skip-validate-pr]

### DIFF
--- a/src/commands/data/export/resume.ts
+++ b/src/commands/data/export/resume.ts
@@ -79,6 +79,11 @@ export default class DataExportResume extends SfCommand<DataExportResumeResult> 
       ms.goto('exporting records', { state: jobInfo.state });
     });
 
+    queryJob.on('error', (err) => {
+      ms.stop('failed');
+      throw err;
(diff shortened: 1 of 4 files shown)


src/commands/data/export/resume.ts
```diff
@@ -79,6 +79,11 @@ export default class DataExportResume extends SfCommand<DataExportResumeResult>
       ms.goto('exporting records', { state: jobInfo.state });
     });
 
+    queryJob.on('error', (err) => {
+      ms.stop('failed');
+      throw err;
+    });
```
cristiand391, line 85: noticed this was missing, it should make sf throw if polling/fetching records in `exportRecords` fails.

mdonnalley (approved):

shetzel (approved):

PR #1092, opened 2024-10-16
Wr/export junction objects

WillieRuemmele:
### What does this PR do?
allows multiple `--query` to `data:export:tree`

@W-5346871@

### DIFF
--- a/src/commands/data/export/tree.ts
+++ b/src/commands/data/export/tree.ts
@@ -11,12 +11,12 @@ import { orgFlags, prefixValidation } from '../../../flags.js';
 import { ExportConfig, runExport } from '../../../export.js';
 import type { DataPlanPart, SObjectTreeFileContents } from '../../../types.js';
 
-export type ExportReturnType = DataPlanPart[] | SObjectTreeFileContents;
+export type ExportTreeResult = DataPlanPart[] | SObjectTreeFileContents;
 
 Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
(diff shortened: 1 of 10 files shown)


src/commands/data/export/tree.ts
```diff
@@ -11,10 +11,12 @@ import { orgFlags, prefixValidation } from '../../../flags.js';
 import { ExportConfig, runExport } from '../../../export.js';
 import type { DataPlanPart, SObjectTreeFileContents } from '../../../types.js';
 
+export type ExportReturnType = DataPlanPart[] | SObjectTreeFileContents;
```
shetzel, line 14: Unfortunately we are not consistent with our naming patterns for command return results.  This one would be quite different so let's go with the most common pattern; "*Result".  E.g., `ExportTreeResult`

shetzel (changes requested):

shetzel (approved):

shetzel:
QA

1. Linked the changes to my CLI
2. imported data to my scratch org (Account, Contact, AccountContactRelation)
3. exported data using multiple queries
4. inspected the data files that were generated ✅ 

Data export command:
```
sf data:export:tree --plan --output-dir junction --query "select AccountId, ContactId from AccountContactRole" --query "Select Id, AccountId, FirstName, LastName from Contact" --query "select Id, ContactId, AccountId from AccountContactRelation where Account.Name != 'We Know Everybody'" --query "select ID, Name from Account where Name != 'Sample Account for Entitlements'"
```

PR #1100, opened 2024-10-23
feat: use new table

mdonnalley:
### What does this PR do?

Use new table

@W-16736186@


cristiand391 (approved):

cristiand391:
**QA notes**:

setup:
running this branch via `./bin/run.js` and old tables via `sf` v2.64.6
imported ~76K  records into the org  by running:
```
sf data import bulk --file test/test-files/data-project/data/bulkUpsertLarge.csv --sobject account --wait 10
```

❌ `data query -q 'select id,name from account'` frozen/consumes a lot of memory

with the default 50K query limit it stays frozen for a few seconds consuming ~5GB, had to kill the process
Screenshot 2024-10-24 at 6 28 09 PM


If I set it to 10K records it finishes after a bit but still see it comsumes ~1GB at the end:
```
SF_ORG_MAX_QUERY_LIMIT=10000 ./bin/run.js data query -q 'select id,name from account' > new.txt
wc -l new.txt
   10005 new.txt
```

on sf I'm able to query 50K and memory usage stayed at ~180MB during the whole execution.
Screenshot 2024-10-24 at 6 26 35 PM

mdonnalley:
@cristiand391 I think https://github.com/oclif/table/pull/32 addresses most of these issues. I've also push a commit to wrap the json objects in the table instead of truncating them

cristiand391:
**QA updates**:

> ❌ data query -q 'select id,name from account' frozen/consumes a lot of memory

❌  the 10K limit for styling works but still see high memory usage even with 5K records if I query more than 2 fields:

Screenshot 2024-10-28 at 11 00 56 AM

```
./bin/run.js data query -q 'select id,name,BillingCountry,BillingState,BillingStreet,Description from account limit 5000' > new.txt
```

Screenshot 2024-10-28 at 11 06 21 AM

> ❌ JSON fields in table are truncated:

✅ fixed

mdonnalley:
@cristiand391 

>the 10K limit for styling works but still see high memory usage even with 5K records if I query more than 2 fields:

I'm seeing 800mb-1gb on 9999k and ~200mb on 5k. I have an env var for changing the limit at which ink is no longer used (`OCLIF_TABLE_LIMIT`) - other than getting rid of ink entirely, there's not another way to decrease the memory usage

PR #1110, opened 2024-10-31
feat: multi-stage-output/new csv flags

cristiand391:
### What does this PR do?

* Adds `--line-ending` flag to bulk commands that take a CSV file as input
* update `sf data delete/upsert bulk` commands to use MSO

@W-16971607@
https://github.com/forcedotcom/cli/issues/3061


src/bulkDataRequestCache.ts
```diff
@@ -64,61 +64,33 @@ export abstract class BulkDataRequestCache extends TTLConfig<TTLConfig.Options,
...
+      if (!entry) {
+        throw messages.createError('error.bulkRequestIdNotFound', [jobIdOrMostRecent]);
+      }
+
+      return {
+        jobInfo: { id: entry.jobId },
+        options: {
+          connection: (await Org.create({ aliasOrUsername: entry.username })).getConnection(),
+        },
+      };
```
cristiand391, line 93: simplified `resolveResumeOptionsFromCache` so that it matches `data import/export`'s cache resolver.
The previous implementation had a few unused properties and one (undocumented?) functionality:
passing a job ID that didn't exist in the cache wouldn't cause an error, the resolver would return it as if it was found in the local cache.
Added code ƒor specific bulk commands using this to avoid breaking changes.

src/bulkIngest.ts
```diff
@@ -37,34 +38,45 @@ type ResumeCommandIDs = 'data import resume' | 'data update resume';
  *
  * It will create the specified bulk ingest job, set up the oclif/MSO stages and return the job info.
  * */
+// eslint-disable-next-line complexity
```
cristiand391, line 41: had to add this after adding the warning/deprecation/validation checks.

src/bulkIngest.ts
```diff
@@ -37,34 +38,45 @@ type ResumeCommandIDs = 'data import resume' | 'data update resume';
...
-    lineEnding = platform() === 'win32' ? 'CRLF' : 'LF',
-    columnDelimiter,
-    file,
-    logFn,
-  } = opts;
+  const { conn, operation, object, lineEnding = platform() === 'win32' ? 'CRLF' : 'LF', file, logFn } = opts;
+
+  // validation
+  if (opts.externalId && opts.operation !== 'upsert') {
+    throw new SfError('External ID is only required for `sf data upsert bulk.');
```
WillieRuemmele, line 64: bulk.` - missing last '`'

src/bulkIngest.ts
```diff
@@ -37,34 +38,45 @@ type ResumeCommandIDs = 'data import resume' | 'data update resume';
...
+  }
+
+  if (opts.verbose && !['delete', 'hardDelete', 'upsert'].includes(opts.operation)) {
+    throw new SfError(
+      'Verbose mode is limited for `sf data delete/upsert bulk` for backwards-compat only and will be removed after March 2025.'
```
WillieRuemmele, line 69: "backwards-compat" - @jshackell-sfdc

src/bulkIngest.ts
```diff
@@ -172,10 +225,11 @@ export async function bulkIngest(opts: {
 export async function bulkIngestResume(opts: {
   cmdId: ResumeCommandIDs;
   stageTitle: string;
-  cache: BulkUpdateRequestCache;
+  cache: BulkUpdateRequestCache | BulkUpsertRequestCache;
   jobIdOrMostRecent: string | boolean;
   jsonEnabled: boolean;
   wait: Duration;
+  warnFn: (message: SfCommand.Warning) => void;
```
WillieRuemmele, line 232: were you matching a pattern here? passing in a logFn/warnFn?

src/commands/data/export/resume.ts
```diff
@@ -53,10 +53,8 @@ export default class DataExportResume extends SfCommand<DataExportResumeResult>
     const queryJob = new QueryJobV2(resumeOpts.options.connection, {
       id: resumeOpts.jobInfo.id,
       pollingOptions: {
-        // both `pollInterval` and `pollTimeout` values are 0 (set in BulkExportRequestCache.resolveResumeOptionsFromCache).
-        // So we set the interval to 5s and timeout to 30s (otherwise jsforce would throw a timeout err if passed 0ms).
-        pollInterval: Math.max(resumeOpts.options.pollingOptions.pollInterval, 5000),
-        pollTimeout: Math.max(resumeOpts.options.pollingOptions.pollTimeout, 30_000),
+        pollInterval: Math.max(5000),
```
WillieRuemmele, line 56: why `Math.max`? just set to 5000?

src/resumeBulkBaseCommand.ts
```diff
@@ -43,7 +43,7 @@ export abstract class ResumeBulkCommand extends SfCommand<BulkResultV2> {
       summary: messages.getMessage('flags.wait.summary'),
       unit: 'minutes',
       min: 0,
-      defaultValue: 0,
+      defaultValue: 5,
```
WillieRuemmele, line 46: this will affect commands that don't currently pass `--wait` and that are expecting async behavior

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---
on a call with Cristian to verify everything
✅ : flag deprecation warnings
✅ : resume an invalidated cache entry
✅ : json error information
✅ : auto-detect Column delims
✅ : using column delimiters 
✅ : line endings

amtrack:
@cristiand391 When importing a CSV file which has **only 1 column** it looks like the auto-detection of the column delimiter fails (and therefore also the import job):

> Creating Account
>
> Error (SfError): Failed to detect column delimiter used in data/Account.csv.
>
> Job failed

I'm able to work around this by using `--column-delimiter=","` explicitly for the files having 1 column only.

Do you think there is a way to make this work without breaking anything?
If not, I think it's OK.

cristiand391:
@amtrack thanks for reporting, fixed it ⬆️ (available in the nightly channel)

PR #1135, opened 2024-11-28
feat(data:query): add `--output-file` flag

cristiand391:
### What does this PR do?

Adds `--output-file` flag to `data query` command.

`data query` has 3 output formats (`csv`, `json` or `human`, can be set via `--result-format`), the new `--output-file` will only work with `csv` and `json` format b/c:
1. those 2 formats are consumable by other tools
2. `human` format uses oclif/table which can't make it write output to a file instead of stdout without some additional work.

expected behavior:
* no `--output-file` -> records printed to stdout
* `--output-file` and `--result-format` csv or json -> records saved to file
* `--output-file` and `--result-format` csv or json and `--json` -> records saved to file, json output includes file path
* `--output-file` and `--result-format human` -> error

See https://salesforce.quip.com/QuS1ABld0leA

@W-17336452@

### DIFF
--- a/messages/reporter.md
+++ b/messages/reporter.md
@@ -1,3 +1,0 @@
-# bulkV2Result
-
-Job %s | Status %s | Records processed %d | Records failed %d
--- a/messages/soql.query.md
+++ b/messages/soql.query.md
@@ -58,7 +58,7 @@ Time to wait for the command to finish, in minutes.
 
 # flags.output-file.summary
 
-File where records are written.
+File where records are written; only CSV and JSON output formats are supported. 
 
 # displayQueryRecordsRetrieved
 
--- a/messages/soql.query.md
+++ b/messages/soql.query.md
@@ -16,9 +16,9 @@ When using --bulk, the command waits 3 minutes by default for the query to compl
 
   <%= config.bin %> <%= command.id %> --query "SELECT Id, Name, Account.Name FROM Contact"
 
-- Read the SOQL query from a file called "query.txt"; the command uses the org with alias "my-scratch":
+- Read the SOQL query from a file called "query.txt" and write the CSV-formatted output to a file; the command uses the org with alias "my-scratch":
 
-  <%= config.bin %> <%= command.id %> --file query.txt --target-org my-scratch
+  <%= config.bin %> <%= command.id %> --file query.txt --output-file output.csv --result-format csv --target-org my-scratch
 
 - Use Tooling API to run a query on the ApexTrigger Tooling API object:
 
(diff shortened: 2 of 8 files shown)


messages/reporter.md
```diff
@@ -1,3 +0,0 @@
-# bulkV2Result
-
-Job %s | Status %s | Records processed %d | Records failed %d
```
cristiand391, line 3: unused msg so I removed the whole file, left over from previous PR where we refactored bulk commands to use oclif/multi-stage-output.

src/commands/data/query.ts
```diff
@@ -81,6 +88,22 @@ export class DataSoqlQueryCommand extends SfCommand<unknown> {
...
+      summary: messages.getMessage('flags.output-file.summary'),
+      relationships: [
+        {
+          type: 'some',
+          flags: [
+            {
+              name: 'result-format',
+              // eslint-disable-next-line @typescript-eslint/require-await
+              when: async (flags): Promise<boolean> =>
+                flags['result-format'] === 'csv' || flags['result-format'] === 'json',
```
cristiand391, line 101: validates `--output-file` is only specified if `--result-format` value is `csv` or `json`.

messages/soql.query.md
```diff
@@ -56,6 +56,10 @@ Include deleted records. By default, deleted records are not returned.
 
 Time to wait for the command to finish, in minutes.
 
+# flags.output-file.summary
+
+File where records are written.
```
jshackell-sfdc, line 61:
```suggestion
File where records are written; only CSV and JSON output formats are supported. 
```

jshackell-sfdc: Okay?  I got that from your PR comments.

jshackell-sfdc: Also, how about we update the second existing example to show how to use this new flag?  For example:

- Read the SOQL query from a file called "query.txt" and write the CSV-formatted output to a file; the command uses the org with alias "my-scratch":

  <%= config.bin %> <%= command.id %> --file query.txt --output-file output.csv --result-format csv --target-org my-scratch

cristiand391: @jshackell-sfdc done, ty!

soridalac (approved):

soridalac:
QA Notes

✅ : throw error using --output-file and --result-format human --json
✅ : throw error using --output-file and --result-format human
✅ : throw error using --output-file without --result-format
✅ : Result with 3 accounts records in json output and url using --output-file and --result-format csv or json and --json
✅ : Result with 3 records written to CSV file using --output-file and --result-format csv or json

PR #1172, opened 2025-01-20
fix(bulk-v1): properly handle +1 job batches

cristiand391:
### What does this PR do?

Fixes `sf force data bulk upsert` leaving unresolved promises when upserting +10K records and waiting for it to complete (`--wait <min>`).
The bug only occurs under the following conditions:
* upserting +10K records (the bug happens when waiting for multiple batches to finish)
* synchronous operation (`--wait <min>`)

You can even modify the batcher limit to something lower like 1K and upsert a 2K records csv to repro:
https://github.com/salesforcecli/plugin-data/blob/6f1445be09e847ab6197cc7fbdb095a6e3c79f69/src/batcher.ts#L23

`batchesCompleted`'s initial value from the caller is `0` here:
https://github.com/salesforcecli/plugin-data/blob/6f1445be09e847ab6197cc7fbdb095a6e3c79f69/src/batcher.ts#L261

but `waitForCompletion` is called once per batch so its value on each call was always 0 (from reading the code it seems it expected the`batchesCompleted` counter to be treated as a global variable).

-------------------

#### The missing `force:data:bulk:*` NUTs
We had NUTs for the `force:data:bulk:*` commands here:
https://github.com/salesforcecli/plugin-data/blob/6f1445be09e847ab6197cc7fbdb095a6e3c79f69/test/commands/force/data/bulk/dataBulk.nut.ts

but we stopped running them around Oct. 2024 when I missed them here:
https://github.com/salesforcecli/plugin-data/pull/1099

The first commit in this PR adds them back to confirm they still work
The second commit adds a test case for large upsert, you can see it exited with code 13:
Screenshot 2025-01-21 at 9 57 27 AM

See:
https://www.stefanjudis.com/today-i-learned/top-level-promise-handling-in-node-js/

The third commit includes the fix, all NUTs pass successfully.

### Testing
1. verify you are on node >= 22
2. create a scratch org
3. try doing a large, sync upsert like this and see it fail: https://github.com/salesforcecli/plugin-data/blob/d60ee7528263d49adbff3a0d7a8e736aa0fbb3da/test/commands/force/data/bulk/dataBulk.nut.ts#L130

Also open the org and verify the bulk job was correctly processed.

https://github.com/forcedotcom/cli/issues/3180
@W-17606312@

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "version": "oclif readme"
   },
   "dependencies": {
-    "@jsforce/jsforce-node": "^3.6.3",
+    "@jsforce/jsforce-node": "^3.6.4",
     "@oclif/multi-stage-output": "^0.8.1",
     "@salesforce/core": "^8.6.1",
     "@salesforce/kit": "^3.2.2",
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "test:nuts:data:search": "nyc mocha \"./test/commands/data/search.nut.ts\" --slow 4500 --timeout 600000 --parallel --jobs 20",
     "test:nuts:data:create": "nyc mocha \"./test/commands/data/create/*.nut.ts\" --slow 4500 --timeout 600000 --parallel --jobs 20",
     "test:nuts:data:bulk-upsert-delete": "nyc mocha \"./test/commands/data/dataBulk.nut.ts\" --slow 4500 --timeout 600000 --parallel --jobs 20",
+    "test:nuts:force:data:bulk-upsert-delete-status": "nyc mocha \"./test/commands/force/data/bulk/dataBulk.nut.ts\" --slow 4500 --timeout 600000 --parallel --jobs 20",
     "test:only": "wireit",
     "version": "oclif readme"
   },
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -114,11 +114,12 @@
     "test:nuts:data:search": "nyc mocha \"./test/commands/data/search.nut.ts\" --slow 4500 --timeout 600000 --parallel --jobs 20",
     "test:nuts:data:create": "nyc mocha \"./test/commands/data/create/*.nut.ts\" --slow 4500 --timeout 600000 --parallel --jobs 20",
     "test:nuts:data:bulk-upsert-delete": "nyc mocha \"./test/commands/data/dataBulk.nut.ts\" --slow 4500 --timeout 600000 --parallel --jobs 20",
+    "test:nuts:force:data:bulk-upsert-delete-status": "nyc mocha \"./test/commands/force/data/bulk/dataBulk.nut.ts\" --slow 4500 --timeout 600000 --parallel --jobs 20",
```
cristiand391, line 117: I missed this one when breaking up NUTs in separate runs here:
https://github.com/salesforcecli/plugin-data/pull/1099

soridalac (approved):

soridalac (approved):

soridalac:
**QA notes**:

✅ exit code return 0 as expected result with node v22, using `sf force data bulk upsert` with `--wait ` upserting +10K records and wait to be complete.

PR #1190, opened 2025-02-19
fix: W-17774391  - create table for data import tree JSON Error

soridalac:
### What does this PR do?
create table for JSON error when using` data import tree`.
@W-17774391@

### DIFF
--- a/src/commands/data/import/tree.ts
+++ b/src/commands/data/import/tree.ts
@@ -5,12 +5,13 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { Messages } from '@salesforce/core';
+import { Messages, SfError } from '@salesforce/core';
 import { SfCommand, Flags } from '@salesforce/sf-plugins-core';
+import { isObject } from '@salesforce/ts-types';
 import { importFromPlan } from '../../../api/data/tree/importPlan.js';
 import { importFromFiles } from '../../../api/data/tree/importFiles.js';
 import { orgFlags } from '../../../flags.js';
-import type { ImportResult } from '../../../api/data/tree/importTypes.js';
+import type { ImportResult, TreeResponse } from '../../../api/data/tree/importTypes.js';
 
 Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 const messages = Messages.loadMessages('@salesforce/plugin-data', 'tree.import');
@@ -18,7 +19,7 @@ const messages = Messages.loadMessages('@salesforce/plugin-data', 'tree.import')
 /**
  * Command that provides data import capability via the SObject Tree Save API.
  */
-export default class Import extends SfCommand<ImportResult[]> {
+export default class Import extends SfCommand<ImportResult[] | undefined> {
   public static readonly summary = messages.getMessage('summary');
   public static readonly description = messages.getMessage('description');
   public static readonly examples = messages.getMessages('examples');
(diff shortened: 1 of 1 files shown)


src/commands/data/import/tree.ts
```diff
@@ -45,23 +46,61 @@ export default class Import extends SfCommand<ImportResult[]> {
     }),
   };
 
-  public async run(): Promise<ImportResult[]> {
+  public async run(): Promise<ImportResult[] | undefined> {
```
cristiand391, line 49: ditto https://github.com/salesforcecli/plugin-data/pull/1190/files#r1964278534

src/commands/data/import/tree.ts
```diff
@@ -45,23 +46,61 @@ export default class Import extends SfCommand<ImportResult[]> {
...
+        columns: [
+          { key: 'refId', name: 'Reference ID' },
+          { key: 'type', name: 'Type' },
+          { key: 'id', name: 'ID' },
+        ],
+        title: 'Import Results',
+      });
+      return results;
+    } catch (err) {
+      const error = err as SfError;
```
cristiand391, line 70:
```suggestion
      const error = err as Error;
```

cristiand391: we can't ensure this is only an `SfError` so casting as `Error` is safer.

src/commands/data/import/tree.ts
```diff
@@ -45,23 +46,61 @@ export default class Import extends SfCommand<ImportResult[]> {
...
+      if (
+        error.cause &&
+        error.cause instanceof Error &&
+        'data' in error.cause &&
+        isObject(error.cause.data) &&
+        'message' in error.cause.data
```
cristiand391, line 76: move all these checks to a small function like this:
https://github.com/salesforcecli/sf-plugins-core/blob/2bc43599313191af6df93cccf70fcfe0082c768f/src/errorHandling.ts#L60

see: https://www.typescriptlang.org/docs/handbook/2/narrowing.html

src/commands/data/import/tree.ts
```diff
@@ -45,23 +46,61 @@ export default class Import extends SfCommand<ImportResult[]> {
...
+      ) {
+        const errorData = JSON.parse(error.cause.data.message as string) as TreeResponse;
```
cristiand391, line 78:
```suggestion
        const errorData = JSON.parse(ensureString(error.cause.data.message)) as TreeResponse
```

src/commands/data/import/tree.ts
```diff
@@ -45,23 +46,61 @@ export default class Import extends SfCommand<ImportResult[]> {
...
+
+          this.table({
+            data: errorResults,
+            columns: [
+              { key: 'referenceId', name: 'Reference ID' },
+              { key: 'StatusCode', name: 'Status Code' },
+              { key: 'Message', name: 'Error Message' },
+              { key: 'fields', name: 'Fields' },
+            ],
+            title: 'import Errors',
```
cristiand391, line 100:
```suggestion
            title: 'Tree import errors',
```

src/commands/data/import/tree.ts
```diff
@@ -45,23 +46,61 @@ export default class Import extends SfCommand<ImportResult[]> {
...
+          });
```
cristiand391, line 101: since we manually report the errors in the table we need to manually set the exit code to 1 here:
https://github.com/salesforcecli/plugin-deploy-retrieve/blob/a45867d06f7f511e973a955d4bab164d6486863d/src/commands/project/delete/source.ts#L306

cristiand391 (approved):

cristiand391:
**QA notes**:
setup:
scratch org using ebikes-lwc project

❌ JSON parse failure instead of error msg on non-sobject tree API error responses:
Screenshot 2025-02-21 at 16 26 58

NOTE: I put an invalid sobject in the plan file to get the err

🟢 table with record errors printed correctly
⚠️ JSON error is still show after table:
Screenshot 2025-02-21 at 16 58 29

🟢 if error doesn't have `cause` set they are correctly re-thrown

PR #1209, opened 2025-03-17
fix: remove deprecations W-18057437

cristiand391:
### What does this PR do?
Removes deprecated functionality for `sf data delete/upsert bulk`:
* Remove record failures from JSON output, should use data bulk results to get failures
* No longer allow resuming sync operations without cache

@W-18057437@


WillieRuemmele (approved):

WillieRuemmele:
✅ : saw warning in `--json` warnings section about data being removed in March '25
✅ : data now removed from json
✅ : warning about resuming async operation displayed when using `--wait`
✅ : no longer able to resume ID created with `--wait`

PR #1221, opened 2025-04-07
W-17774659: edit help messages

jshackell-sfdc:
### What does this PR do?

Quick edit of changes to messages due to removing the --async flag.

@W-17774659@


EstebanRomero84 (approved):
It is much better. Thanks Juliet!

PR #1222, opened 2025-04-09
W-17774659: deprecate --async flag on async-by-default commands

EstebanRomero84:
### What does this PR do?
Deprecates the --async flag in the following commands:

1. data delete bulk
2. data export bulk
3. data import bulk
4. data query
5. data update bulk
6. data upset bulk

These commands already run asynchronously by default.
[@W-17774659@ ](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000029IIaZYAW/view)


WillieRuemmele:
### QA Notes
---
✅ : quick sanity check by using deprecated flags=>deprecation messages in stdout/json

PR #1474, opened 2026-07-16
chore: generated schemas folder (W-23365673)

jfeingold35:
### What does this PR do?

It bumps the dependency on `@oclif/plugin-command-snapshot` to a new version that is capable of generating a `schemas` folder, and generates that folder.

@W-23365673@

### DIFF
--- a/src/commands/data/import/tree.ts
+++ b/src/commands/data/import/tree.ts
@@ -20,7 +20,7 @@ import { ensureString, isObject } from '@salesforce/ts-types';
 import { importFromPlan } from '../../../api/data/tree/importPlan.js';
 import { importFromFiles } from '../../../api/data/tree/importFiles.js';
 import { orgFlags } from '../../../flags.js';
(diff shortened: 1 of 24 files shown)


src/commands/data/import/tree.ts
```diff
@@ -20,15 +20,15 @@ import { ensureString, isObject } from '@salesforce/ts-types';
...
+import type { ImportResults, TreeResponse } from '../../../api/data/tree/importTypes.js';
 
 Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 const messages = Messages.loadMessages('@salesforce/plugin-data', 'tree.import');
 
 /**
  * Command that provides data import capability via the SObject Tree Save API.
  */
-export default class Import extends SfCommand<ImportResult[]> {
+export default class Import extends SfCommand<ImportResults> {
```
jfeingold35, line 31: The schema generator didn't like having a return type be an array of X, so I just defined a type that is itself an array and that is working fine.

iowillhoit (approved):

